PR #81, opened 2018-08-20
Ignore extra keys in textDocument/publishDiagnostics

pdixon:
clangd recently started including `:category` in it's diagnostic notification. According to the comments in the [review](https://reviews.llvm.org/D50571) of the change that introduced this VSCode is fine with this extension.


pdixon:
Cool. Thank you.

PR #103, opened 2018-09-08
Close #52: Use entire line as xref summary when possible

### LINKED ISSUE #52: Show context of references
eglot:
eglot
lsp-mode:
lsp-mode

Eglot also displays the definition as a reference.

mkcms:
This commit makes xref items show entire line where a reference was found.
For performance reasons, I made it only show that line when a buffer visiting the reference was found, although I didn't really benchmark.

### DIFF
--- a/eglot.el
+++ b/eglot.el
@@ -1272,13 +1272,30 @@ DUMMY is ignored."
 (advice-add 'xref-find-definitions :after #'eglot--xref-reset-known-symbols)
 (advice-add 'xref-find-references :after #'eglot--xref-reset-known-symbols)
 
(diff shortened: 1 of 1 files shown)


cmm:
1. The CI failure seems transient and unrelated to the PR's content

2. `lsp-mode` shows xref entry context for both already-opened and unopened files (they use `with-temp-file` or something for the latter) -- if there are any performance complaints because of this, I don't know of any.  Given the amount of RAM on today's normal developer machine and how modern FS caches work, I don't see why that would adversely impact the interactive experience in any way

mkcms:
I'm closing this PR in favor of #127.

PR #111, opened 2018-09-12
Truncate hovering information in echo area

fbergroth:
Currently, hovering symbols with multiple lines of hover-info makes the echo area jump around a bit much.

I'm not sure if this should be customizable, we could check `eldoc-echo-area-use-multiline-p`, but I'd like this behavior only for hovering, and not for signatureHelp.

### DIFF
--- a/eglot.el
+++ b/eglot.el
@@ -1548,7 +1548,8 @@ If SKIP-SIGNATURE, don't try to send textDocument/signatureHelp."
                           (when-let (info (and contents
                                                (eglot--hover-info contents
                                                                   range)))
-                            (eldoc-message info)))))
+                            (let ((message-truncate-lines t))
+                              (eldoc-message info))))))
          :deferred :textDocument/hover))
       (when (eglot--server-capable :documentHighlightProvider)
         (jsonrpc-async-request


fbergroth:
@joaotavora are you interested in this patch, or should I close it?

joaotavora:
I don't know yet, but leave it open.

fbergroth:
Okay, so the current behavior keeps annoying me. I'm not sure if it's a problem with other servers, but with python I think it is.

When you hover a module, you get the entire docstring of the module printed in the echo area, which is sometimes rather long. If this docstring doesn't fit, you can only see the last part of it. When navigating through the code, the echo area jumps a lot, which is distracting.

If I want to see the entire documentation, I prefer to call `eglot-help-at-point`.

How about adding a variable
```
(defcustom eglot-hover-display-function #'eldoc-message)
```

And replacing ` (eldoc-message info)` with `(funcall eglot-hover-display-function info)`?

Then I can have in my own config (or the function may be suitable to have in eglot)

```
(defun eglot-eldoc-message-first-line (text)
  (let* ((end (min (1- (window-width (minibuffer-window)))
                   (or (string-match-p "\n" text)
                       (length text))))
         (msg (substring text 0 end)))
    (eldoc-message msg)))
```

and `(setq-local eglot-hover-display-function #'eglot-eldoc-message-first-line)` in my python buffers.

It would also be easier to display hover information with tooltips, if some users prefer that.

mkcms:
@fbergroth Take a look here, maybe it will be helpful. https://github.com/joaotavora/eglot/issues/112#issuecomment-446039123

> How about adding a variable
> 
> ```
> (defcustom eglot-hover-display-function #'eldoc-message)
> ```

Looks like there's interest in other display mechanisms. I personally have `:hoverProvider` disabled all the time. Maybe we can consider adding this variable, or customizing this behavior some other way.

joaotavora:
Hi @fbergroth, sorry for being so late to the party. I had a look at the problem and this indeed as @mkcms says, it looks like something that could be made with `eldoc-message-function`. Also, instead of setting it globally  you could/should use `add-function`:

```
(add-function :around (local 'eldoc-message-function)
              (lambda (original-fn &rest message-args)
                ;; massage message-args to cut off the bits that annoy you
                (apply original-fn message-args)))
```

Furthermore, you probably want to do this only when Eglot is active, right?  For now you probably need to use `eglot--managed-mode-hook`, an internal hook, but see https://github.com/joaotavora/eglot/issues/182.

Also, can you post a screenshot or describe in some more detail exactly what the bother is with the jumping eldoc?  Perhaps you could do something like @casouri does in https://github.com/casouri/eglot-doc, but simpler, and completely based on `eldoc-message-function` instead of a large extension that repeats some code in Eglot (though I have not tested that extension).

fbergroth:
> Also, can you post a screenshot or describe in some more detail exactly what the bother is with the jumping eldoc?

Sure, here's a gif demonstrating the hovering docs being cut off. On a larger display, the echo area takes up a lot more space, and is appearing/disappearing ("jumping") a lot as you navigate through the code.

x

fbergroth:
> Also, instead of setting it globally you could/should use add-function:

Really? `eldoc-message-function` is a variable. Isn't `(setq-local eldoc-message-function my-eldoc-message-first-line)` in i.e. python buffers more appropriate?

joaotavora:
> Really? eldoc-message-function is a variable. 

Yes really :-)

In lisp, we rather look at it as a symbol which has a variable binding to some value.  In this case Emacs guarantees that type of that value is a function (named or unnamed, doesn't matter).  So you can use `add-function`.

Try this in your python mode buffers
```lisp
(add-function :around (local 'eldoc-message-function)
              (lambda (oldfun format-string &rest args)
                (if-let (frame (and
                                (string-match "\n" format-string)
                                (cdr (assoc "*eldoc frame*" (make-frame-names-alist)))))
                  (with-selected-frame frame
                    (delete-other-windows)
                    (with-current-buffer (get-buffer-create " *eldoc big doc*")
                      (erase-buffer)
                      (insert (apply #'format format-string args))
                      (switch-to-buffer (current-buffer))))
                  (apply oldfun args)))
              '((name . joaot/use-separate-frame-maybe)))
```

Won't do anything unless the frame exists, which you can make with `(make-frame '((name . "*eldoc frame*")))` 

> Isn't (setq-local eldoc-message-function my-eldoc-message-first-line) in i.e. python buffers more appropriate?

If you are using `:around`, it's just a stylistic difference I guess (though I prefer my style because it shows up in the `C-h o` of `eldoc-message-function`).

If you are using `:before-until` it's slightly more idiomatic:

```lisp
(add-function :before-until (local 'eldoc-message-function)
              (lambda (format-string &rest args)
                (when-let (frame (and
                                format-string
                                ;; (string-match "\n" format-string)
                                (cdr (assoc "*eldoc frame*" (make-frame-names-alist)))))
                  (with-selected-frame frame
                    (delete-other-windows)
                    (with-current-buffer (get-buffer-create " *eldoc big doc*")
                      (erase-buffer)
                      (save-excursion (insert (apply #'format format-string args)))
                      (switch-to-buffer (current-buffer))))))
              '((name . joaot/use-separate-frame-maybe)))
```

Anyway:

* You should still decide what happens if you don't have the separate frame and the message is still too large (you could just ignore it).
* This solution is completion eglot-independent, but you might want to do it in `eglot--managed-mode-hook` (adding and removing it after eglot stops managing).
* But the same argument goes for making it global, especially since it would handle other eldoc "misbehavers", not just LSP-based ones. In that case just remove the `local` from the `add-function`.

fbergroth:
Thanks for the detailed answer. I learned something new :-)

Just to make it clear, I don't want a frame that displays documentation during hover. I would find that as distracting as eldoc being hammered with pages of text. I do think that eglot should support a lighter variant of hover documentation, instead of passing the entire thing (my suggestion is still to only display the first line with `eldoc-message`). But now I know how to fix this for myself, so I'll leave it there.

joaotavora:
> Thanks for the detailed answer. I learned something new :-)

You're welcome.

> I do think that eglot should support a lighter variant of hover documentation, instead of passing the entire thing

But am I mistaken in thinking this is something you would want for any eldoc thing, and not necessarily related to eglot?  If I'm not, then you should request that option of eldoc, i.e. Emacs itself.  Use `M-x report-emacs-bug` for that, and put me in Cc.

fbergroth:
I've only encountered this in eglot, so I'm not sure if it applies elsewhere. Also as I mentioned in my first post, I only want this for hovering, not for signatureHelp, so some change would have to bemade in eglot. Maybe it's not worth fixing unless more people find it distracting.

joaotavora:
> I've only encountered this in eglot, so I'm not sure if it applies elsewhere.

By your description, it applies to every user of eldoc that uses it to issue longer-than-one-line messages. 

> for signatureHelp, so some change would have to bemade in eglot.

So if a signatureHelp is longer than one line that's OK? 

> Maybe it's not worth fixing unless more people find it distracting.

You can M-x report-emacs-bug anyway.

fbergroth:
> By your description, it applies to every user of eldoc that uses it to issue longer-than-one-line messages.

I guess the modes that I've previously used only sent short text to eldoc, or it was customizable (i.e. `anaconda-mode-eldoc-as-single-line`).

> So if a signatureHelp is longer than one line that's OK?

Yeah, for most cases I just see two lines here, the first being the function signature, and the latter information about the parameter I'm about to type. I think that's sensible.

casouri:
I added a commit to eldoc-box(eglot-doc) so you can chose to leave one line messages in minibuffer and display multi-line messages in childframe. Maybe that's something you want?

https://github.com/casouri/eldoc-box/commit/194ff4136ac463bd79ca8d35b72851248372b89a

fbergroth:
@casouri eldoc-box looks very nice! For some reason child frames causes my emacs to flicker, but I'll probably dig into why, since I'd like to use eldoc-box.

I don't think I'd use `eldoc-box-only-multi-line` though. I'd like to know ahead of time where text pops up, and where to look for it :-)

joaotavora:
Yes, I tried it too.  It's a very good first start.  It's kind of unusably slow on Windows, but that's probably fixable.  

Why does one have to close the  new frame everytime?

There are some mentions of `lsp` there. Is  all the code yours?

casouri:
>For some reason child frames causes my emacs to flicker

Does [this](https://github.com/casouri/eldoc-box/commit/f0c99152d21f6bf4694064a6e204cd4a8a229b07) fix it?

> Why does one have to close the new frame every time?

That can be changed to `make-frame-invisible`, I'll try that.

> There are some mentions of lsp there. Is all the code yours?

I went to lsp-ui-doc.el and copied a bunch of code. Is there anything I need to do?

joaotavora:
> That can be changed to make-frame-invisible, I'll try that.

Cool.

> I went to lsp-ui-doc.el and copied a bunch of code. Is there anything I need to do?

Well IANAL, but there's a bit there at the top that says that the code is your copyright...

casouri:
Oops! I'll remove that... That's embarrassing...

joaotavora:
Don't worry too much, just add your name to the old copyright notice. We
can evaluate later if that's a problem for integrating into ELPA or
something like that

On Dec 11, 2018 20:44, "Yuan Fu" <notifications@github.com> wrote:

Oops! I'll remove that... That's embarrassing...

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/joaotavora/eglot/pull/111#issuecomment-446355020>, or mute
the thread

PR #145, opened 2018-11-04
Make imenu hierarchical

mkcms:
* eglot.el (eglot-imenu): Use :containerName to build a nested imenu
  index alist.


MaskRay:
`textDocument/documentSymbol` may return `DocumentSymbol[]` (hierarchical document symbol) which you might also want to take a look.

MaskRay:
ccls supports it.
Some typescript or javascript language server may support it.
There appears to be a pending clangd revision to support it.

PR #156, opened 2018-11-18
WIP: pcase macro

### LINKED ISSUE #144: Microsoft's python language server sending non-standard keys
I've tried to use eglot with microsoft's [python language server](https://github.com/Microsoft/python-language-server) and the following error was raised:

```
Debugger entered--Lisp error: (error "Keyword argument :_version not one of (:contents :range)")
  signal(error ("Keyword argument :_version not one of (:contents :range)"))
  error("Keyword argument %s not one of (:contents :range)" :_version)
  #f(compiled-function (&rest rest) #<bytecode 0x430a9b0d>)(:contents (:kind "plaintext" :value "") :range nil :_version nil :_typeNames nil)
  apply(#f(compiled-function (&rest rest) #<bytecode 0x430a9b0d>) (:contents (:kind "plaintext" :value "") :range nil :_version nil :_typeNames nil))
  #f(compiled-function (jsonrpc-lambda-elem13) #<bytecode 0x407f268d>)((:contents (:kind "plaintext" :value "") :range nil :_version nil :_typeNames nil))
  jsonrpc-connection-receive(#<eglot-pyls eglot-pyls> (:jsonrpc "2.0" :id 25 :result (:contents (:kind "plaintext" :value "") :range nil :_version nil :_typeNames nil)))
  jsonrpc--process-filter(#<process EGLOT (Desktop/python-mode)> "Content-Length: 128\015\n\015\n{\n  \"jsonrpc\": \"2.0\",\n  \"id\": 24,\n  \"result\": {\n    \"signatures\": null,\n    \"activeSignature\": 0,\n    \"activeParameter\": 0\n  }\n}Content-Length: 187\015\n\015\n{\n  \"jsonrpc\": \"2.0\",\n  \"id\": 25,\n  \"result\": {\n    \"contents\": {\n      \"kind\": \"plaintext\",\n      \"value\": \"\"\n    },\n    \"range\": null,\n    \"_version\": null,\n    \"_typeNames\": null\n  }\n}")
```

I build the python language server following the [guide](https://github.com/Microsoft/python-language-server/blob/master/CONTRIBUTING.md#setup) and copy the `languageServer` directory to my emacs directory.

The server program I defined:

```emacs-lisp
(setq pyls-command
      (expand-file-name
       "languageServer/Microsoft.Python.LanguageServer"
       user-emacs-directory))
(defclass eglot-pyls (eglot-lsp-server) ()
  :documentation
  "Microsoft's Python Language Server.")

(cl-defmethod eglot-initialization-options ((server eglot-pyls))
  "Passes through required pyls initialization options."
  `((:interpreter
     (:properties
      (:UseDefaultDatabase t :Version "3.7")))))

(add-to-list
 'eglot-server-programs
 `(python-mode eglot-pyls ,pyls-command))
```

The python code I used for testing:
```python
# ~/Desktop/test.py
import os

os.
```

<details>
<summary>Eglot events logs</summary>
<pre><code>
client-request (id:1) Fri Nov  2 15:50:29 2018:
(:jsonrpc "2.0" :id 1 :method "initialize" :params
          (:processId 33063 :rootPath "/Users/galeo/Desktop/" :rootUri "file:///Users/galeo/Desktop/" :initializationOptions
                      ((:interpreter
                        (:properties
                         (:UseDefaultDatabase t :Version 3.6))))
                      :capabilities
                      (:workspace
                       (:applyEdit t :executeCommand
                                   (:dynamicRegistration :json-false)
                                   :workspaceEdit
                                   (:documentChanges :json-false)
                                   :didChangeWatchedFiles
                                   (:dynamicRegistration t)
                                   :symbol
                                   (:dynamicRegistration :json-false))
                       :textDocument
                       (:synchronization
                        (:dynamicRegistration :json-false :willSave t :willSaveWaitUntil t :didSave t)
                        :completion
                        (:dynamicRegistration :json-false :completionItem
                                              (:snippetSupport t))
                        :hover
                        (:dynamicRegistration :json-false)
                        :signatureHelp
                        (:dynamicRegistration :json-false)
                        :references
                        (:dynamicRegistration :json-false)
                        :definition
                        (:dynamicRegistration :json-false)
                        :documentSymbol
                        (:dynamicRegistration :json-false)
                        :documentHighlight
                        (:dynamicRegistration :json-false)
                        :codeAction
                        (:dynamicRegistration :json-false)
                        :formatting
                        (:dynamicRegistration :json-false)
                        :rangeFormatting
                        (:dynamicRegistration :json-false)
                        :rename
                        (:dynamicRegistration :json-false)
                        :publishDiagnostics
                        (:relatedInformation :json-false))
                       :experimental nil)))

server-notification Fri Nov  2 15:50:30 2018:
(:jsonrpc "2.0" :method "window/logMessage" :params
          (:type 0 :message "Microsoft Python Language Server version 0.1.50.0"))

server-notification Fri Nov  2 15:50:30 2018:
(:jsonrpc "2.0" :method "window/logMessage" :params
          (:type 0 :message "Initializing for generic interpreter"))

server-reply (id:1) Fri Nov  2 15:50:30 2018:
(:jsonrpc "2.0" :id 1 :result
          (:capabilities
           (:textDocumentSync
            (:openClose t :change 2 :willSave :json-false :willSaveWaitUntil :json-false :save nil)
            :hoverProvider t :completionProvider
            (:resolveProvider :json-false :triggerCharacters
                              ["."])
            :signatureHelpProvider
            (:triggerCharacters
             ["(,)"])
            :definitionProvider t :referencesProvider t :documentHighlightProvider :json-false :documentSymbolProvider t :workspaceSymbolProvider t :codeActionProvider :json-false :codeLensProvider nil :documentFormattingProvider :json-false :documentRangeFormattingProvider :json-false :documentOnTypeFormattingProvider
            (:firstTriggerCharacter "\n" :moreTriggerCharacter
                                    [";" ":"])
            :renameProvider t :documentLinkProvider nil :executeCommandProvider nil :experimental nil)))

client-notification Fri Nov  2 15:50:30 2018:
(:jsonrpc "2.0" :method "initialized" :params
          (:__dummy__ t))

client-notification Fri Nov  2 15:50:30 2018:
(:jsonrpc "2.0" :method "textDocument/didOpen" :params
          (:textDocument
           (:uri "file:///Users/galeo/Desktop/test.py" :version 0 :languageId "python" :text "import os\n\nos.path.join\n")))

client-notification Fri Nov  2 15:50:30 2018:
(:jsonrpc "2.0" :method "workspace/didChangeConfiguration" :params
          (:settings nil))

server-notification Fri Nov  2 15:50:30 2018:
(:jsonrpc "2.0" :method "python/languageServerStarted" :params
          [])

server-notification Fri Nov  2 15:50:30 2018:
(:jsonrpc "2.0" :method "textDocument/publishDiagnostics" :params
          (:uri "file:///Users/galeo/Desktop/test.py" :diagnostics
                []))

client-request (id:2) Fri Nov  2 15:50:35 2018:
(:jsonrpc "2.0" :id 2 :method "textDocument/signatureHelp" :params
          (:textDocument
           (:uri "file:///Users/galeo/Desktop/test.py")
           :position
           (:line 2 :character 12)))

client-request (id:3) Fri Nov  2 15:50:35 2018:
(:jsonrpc "2.0" :id 3 :method "textDocument/hover" :params
          (:textDocument
           (:uri "file:///Users/galeo/Desktop/test.py")
           :position
           (:line 2 :character 12)))

server-reply (id:2) Fri Nov  2 15:50:35 2018:
(:jsonrpc "2.0" :id 2 :result
          (:signatures nil :activeSignature 0 :activeParameter 0))

server-reply (id:3) Fri Nov  2 15:50:35 2018:
(:jsonrpc "2.0" :id 3 :result
          (:contents
           (:kind "plaintext" :value "")
           :range nil :_version nil :_typeNames nil))
</code></pre>
</details>

mkcms:
Previously we were strict about the structure and contents of LSP
objects.  This led to errors when certain servers sent messages with
fields not defined in the specification (because those fields are
used by other clients, or the protocol evolved and added some fields
and we never upgraded).

This commit makes eglot tolerate unknown keys in all incoming
messages.

* eglot.el (eglot--destructuring-bind): New macro.
(eglot-handle-notification):
(eglot--register-unregister):
(eglot--xref-make):
(xref-backend-apropos):
(eglot-completion-at-point):
(eglot--sig-info):
(eglot-help-at-point):
(eglot--apply-workspace-edit):
(eglot-code-actions):
(eglot--register-workspace/didChangeWatchedFiles): Use it instead of
`cl-destructuring-bind`.

### DIFF
--- a/eglot-tests.el
+++ b/eglot-tests.el
@@ -636,6 +636,40 @@ Pass TIMEOUT to `eglot--with-timeout'."
        (eglot--dbind FooObject (&key foo bar) `(:foo "foo" :baz bargh)
          (cons foo bar))))))
 
+(ert-deftest eglot-pcase ()
+  (should
+   (equal nil
+          (let ((eglot-strict-mode nil))
+            (pcase nil
+              ((eglot--object (:foo) () (foo)) foo)))))
+  (should
+   (equal '("foo" . "bar")
+          (let ((eglot-strict-mode nil))
+            (pcase `(:foo "foo" :bar "bar")
+              ((eglot--object (:foo) (:bar) (foo bar)) (cons foo bar))))))
+  (should
+   (equal '("foo" . "bar")
+          (let ((eglot-strict-mode nil))
+            (pcase `(:foo "foo" :bar "bar" :fotrix bargh)
+              ((eglot--object (:foo) (:bar) (foo bar)) (cons foo bar))))))
+  (should
+   (equal nil
+          (let ((eglot-strict-mode '(disallow-non-standard-keys)))
+            (pcase `(:foo "foo" :bar "bar" :fotrix bargh)
+              ((eglot--object (:foo) (:bar) (foo bar)) (cons foo bar))))))
+  (should
+   (equal 'bargh
+          (let ((eglot-strict-mode '(disallow-non-standard-keys)))
+            (pcase `(:foo "foo" :bar "bar" :fotrix bargh)
+              ((eglot--object (:foo) (:bar) (foo bar)) (cons foo bar))
+              ((eglot--object (:fotrix) (:foo :bar) (fotrix)) fotrix)))))
+  (should
+   (equal "bar"
+          (let ((eglot-strict-mode '(disallow-non-standard-keys)))
+            (pcase `(:foo "foo" :bar "bar")
+              ((eglot--object (:foo :other) () (foo)) foo)
+              ((eglot--object (:foo) (:bar) (bar)) bar))))))
+
 (provide 'eglot-tests)
 ;;; eglot-tests.el ends here
 
--- a/eglot.el
+++ b/eglot.el
@@ -201,19 +201,6 @@ let the buffer grow forever."
 
 ;;; Message verification helpers
 ;;;
-(defvar eglot--lsp-interface-alist `()
-  "Alist (INTERFACE-NAME . INTERFACE) of known external LSP interfaces.
-
-INTERFACE-NAME is a symbol designated by the spec as \"export
-interface\".  INTERFACE is a list (REQUIRED OPTIONAL) where
-REQUIRED and OPTIONAL are lists of keyword symbols designating
-field names that must be, or may be, respectively, present in a
-message adhering to that interface.
-
-Here's what an element of this alist might look like:
-
-    (CreateFile . ((:kind :uri) (:options)))")
-
 (defvar eglot-strict-mode '()
   "How strictly Eglot vetoes LSP messages from server.
 
@@ -229,6 +216,43 @@ is raised if any required fields are missing from the message.
 If the list is empty, any non-standard fields sent by the server
 and missing required fields are accepted (which may or may not
 cause problems in Eglot's functioning later on).")
+
+(eval-and-compile
+  (defvar eglot--lsp-interface-alist
+    `(
+      (CreateFile . ((:kind :uri) (:options)))
+      )
+    "Alist (INTERFACE-NAME . INTERFACE) of known external LSP interfaces.
+
+INTERFACE-NAME is a symbol designated by the spec as \"export
+interface\".  INTERFACE is a list (REQUIRED OPTIONAL) where
+REQUIRED and OPTIONAL are lists of keyword symbols designating
+field names that must be, or may be, respectively, present in a
+message adhering to that interface.
+
+Here's what an element of this alist might look like:
+
+    (CreateFile . ((:kind :uri) (:options)))")
+
+  (pcase-defmacro eglot--object (required-keys optional-keys bindings)
+    (let ((expval (make-symbol "eglot--pcase-expval")))
+      `(and ,expval (pred listp)
+            ,@(cl-loop for key in (append required-keys optional-keys)
+                       for varname = (intern (substring (symbol-name key) 1))
+                       when (memq key required-keys)
+                       collect `(guard (plist-member ,expval ,key))
+                       when (memq varname bindings)
+                       collect `(let ,varname (plist-get ,expval ,key)))
+            (guard (or (not (memq 'disallow-non-standard-keys eglot-strict-mode))
+                       (not (cl-loop
+                             with allowed = '(,@required-keys ,@optional-keys)
+                             for (key _val) on ,expval by #'cddr
+                             unless (memq key allowed) return t)))))))
+
+  (pcase-dolist (`(,name ,required ,optional) eglot--lsp-interface-alist)
+    (eval `(pcase-defmacro ,(intern (format "eglot-%s" name)) (bindings)
+             ,(format "Matches if EXPVAL is LSP %s Object, as a plist." name)
+             `(eglot--object ,',required ,',optional ,bindings)))))
 
 (defun eglot--call-with-interface (interface object fn)
   "Call FN, but first check that OBJECT conforms to INTERFACE.


mkcms:
@joaotavora So is this the right approach? Do we want to always allow other keys? I believe that we should - accept all but use the minimum required. 

You mentioned [here](https://github.com/joaotavora/eglot/pull/138#issuecomment-434280639) that we could add `eglot--dbind` which works like `cl-destructuring-bind` but accepts a type argument. I don't really understand this. Would it be a symbol whose value is a type specification? This seems like a lot of work to maintain a list of all LSP types in eglot and IMO reduces 'locality' (when reading the code, you need to jump around the file to know what the keys are)

joaotavora:
> @joaotavora So is this the right approach? Do we want to always allow other keys? I believe that we should - accept all but use the minimum required.

We want to allow-other-keys _sometimes_ right? We want to decide that at runtime, if possible (or, if it's very slow, at compile-time). Hiding it in a macro such as the one you wrote it's trivial to implement.

Regarding the `type` argument, it could be a symbol like `ResponseMessage` or `CreateFileOptions` or in general anything the specs describes as `export interface`. In a future more sophisticated version of that macro, we can check in runtime if the message being destructured exactly matches the format that some version of the spec defines for that type.

But in a very first version, it's OK for users of the `eglot--dbind` macro and `eglot--lambda` macro to just pass `nil` for that argument, meaning something like "I don't want any validation yet".  But later this might come in handy. Don't worry I can provide that part.

joaotavora:
@mkcms I've pushed a commit that doesn't yet solve the problem but implements some of my ideas.  You can try out the new macros for yourself and start using them instead of `jsonrpc-lambda` and `cl-destructuring-bind`.

In the future, we might want to add more macros on top of those, a "destructuring case" comes to mind.  

This should hopefully make it easier to track back to the spec from the lisp code in eglot.

mkcms:
@joaotavora 
Nice, I will test those macros later. From what I see, one thing that I think would be nice to have, is multiple bindings in one call to `eglot--dbind`, like in `pcase`. E.g. if one matching fails, try another and then signal an error if that fails.
This would for example simplify solutions to #164 (check if an object is a `Command` or a `CodeAction`).

I also wrote some of those macros for fun. I didn't really pay attention to performance or readability though, they're just a proof-of-concept. Also, options to control strict checking aren't implemented yet.

With `pcase`:
```elisp
(pcase-defmacro lsp--Object (fields)
  `(and (pred listp) it
        ,@(mapcar (lambda (field) `(guard (plist-member it ,field)))
                 (butlast fields))
        ,@(mapcar (lambda (field) 
                    `(let ,field
                       (plist-get it ,(intern (format ":%s" field)))))
                  (car (last fields)))))

(pcase-defmacro lsp-TextEdit (symbols)
  `(lsp--Object (:range :newText ,symbols)))

(pcase-defmacro lsp-WorkspaceEdit (symbols)
  `(lsp--Object (:changes :documentChanges ,symbols)))

;; Example
(let ((response '(:range (:start (:character 1 :line 1)
                                 :end (:character 2 :line 1))
                         :newText "x")))
  (pcase response
    ((lsp-WorkspaceEdit (changes documentChanges))
     (format "WorkspaceEdit with changes %S" changes))
    ((lsp-TextEdit (newText range)) 
     (format "TextEdit with text %S" newText))))
```

And a custom macro:

```elisp
(defmacro eglot--dbind (exp &rest clauses)
  (declare (indent 1) (debug (form &rest (sexp sexp body))))
  (let ((expval (make-symbol "eglot--dbind-expval")))
    `(let ((,expval ,exp))
       (cond ,@(mapcar (pcase-lambda (`(,type ,args . ,body))
                         (cond ((eq type 'string)
                                `((stringp ,expval) (let ((,args ,expval)) ,@body)))
                               (t
                                `((cl-every (apply-partially #'plist-member ,expval)
                                            (get ',type 'eglot--destructuring))
                                  (cl-destructuring-bind (&key ,@args) ,expval
                                    ,@body)))))
                       clauses)
             (t (error "No match"))))))

(put 'TextEdit 'eglot--destructuring '(:range :newText))
(put 'WorkspaceEdit 'eglot--destructuring '(:changes :documentChanges))

(eglot--dbind (list :range '(:start (:line 1 :character 1)
                                    (:end (:line 1 :character 5)))
                    :newText "asdf")
  (string x (message "got a string %s" x))
  (WorkspaceEdit (changes documentChanges)
                 (message "got WorkspaceEdit with changes %s" changes))
  (TextEdit (range newText)
            (message "got TextEdit with newText %s" newText)))
```

joaotavora:
> From what I see, one thing that I think would be nice to have, is multiple bindings in one call to eglot--dbind, like in pcase. E.g. if one matching fails, try another and then signal an error if that fails.

This is precisely what the `eglot--dcase` macro shall do. It can make use of `eglot--dbind`.

> I also wrote some of those macros for fun. I didn't really pay attention to performance or readability though, they're just a proof-of-concept. 

These look very good. Perhaps they can substitute my versions, since their API seems to be similar. But I don't have time now to review them now.

Anyway, to fix #144 in the short-term, I think it _won't_ be wasted effort to replace most `cl-destructuring-bind` and `jsonrpc-lambda` with `eglot--dbind` and `eglot--lambda`.  Unless you think the API of my macros isn't good for some reason.

joaotavora:
The `pcase-*` macrology is definitely very good and readable.  It can probably be enhanced to honor `eglot-strict-mode`.

mkcms:
>  It can probably be enhanced to honor `eglot-strict-mode`.

I will try to implement that.

mkcms:
@joaotavora 
I pushed a commit which implements the `pcase` macro.
It's messy... please take a look if you can.
Right now, the macro honors only `disallow-non-standard-keys` of `eglot-strict-mode`  by not matching lists that have extra keys. `enforce-required-keys` is used by default, because the required keys is how the macro knows whether the object was actually matched.
I also modified `eglot--dbind` to use `pcase`. I'm not sure if it's good.

mkcms:
For now, I brought back the old `eglot--dbind` and `eglot--call-with-interface`, because I realised I broke the old behavior.

joaotavora:
This is better some in a separate branch, by the way...

leungbk:
@joaotavora @mkcms Is there still interest in this? I wrote something similar in `lsp-mode` at https://github.com/emacs-lsp/lsp-mode/commit/6bf62bd02da0500df8d638203326ed1d1cbb7521, though I hadn't considered allowing unknown keys. I wouldn't mind helping to wrap this up, if no one has the time to work on it.

skangas:
@leungbk Given the lack of reply in 6 months, I think it's fair to say that any work you would want to do here is very welcome. Just open a new PR with your changes. Thanks in advance!

PR #161, opened 2018-11-20
Add support for R's languageserver

jabranham:
I've been playing around with it a bit and it seems to work pretty well out of the box. R's language server is still pretty young though so a lot of things aren't implemented yet.


mkcms:
Thanks. Merged with slightly improved commit message.

Issue #129, opened 2018-10-14
Use with projectile


paopaol:
i was use emacs 26.
and i place a .projectile file in my root-directory.then i open a cpp file which in my sub-directory. but the eglot use the file current direcotory as the root direcotry in lsp server init param.

lsp-mode works fine. 
how can i fix this problem. thanks

joaotavora:
Eglot doesn't use projectile, it uses Emacs's built-in `project.el`.  See the documentation for `project-find-functions`. By default it searches for a `.git` directory, but your could add a function that searches for `.projectile` files. 

This is entirely untested, it's possible you need to add more functions.

```
(defun joaot/find-projectile-project ()
  (let ((probe (locate-dominating-file default-directory ".projectile")))
    (when probe `(projectile . ,probe))))

(add-hook 'project-find-functions 'joaot/find-projectile-project 'append)
```

bbatsov:
@joaotavora Given how popular Projectile is, it seems to me that you can at least add a check whether it's present and use it's project resolution logic. It's not like I can't add `project.el` integration to Projectile, but I really don't get what's the issue with adding some conditional support for Projectile given how simple `eglot`'s usage of `project.el` is.

I get that project.el is built-in, but having something built-in doesn't necessarily make it the one true way, right? 😉 

Obviously I'm a bit biased, but the "wontfix" tag was certainly disappointing to me.

bbatsov:
I just remembered someone actually brought this up on Projectile's side as well - here's the conversation for reference https://github.com/bbatsov/projectile/issues/1282

joaotavora:
[ @dgutov, sorry to bother you again, but I think maybe you might want to add two cents to this, given you are the project.el maintainer ]

>  It's not like I can't add project.el integration to Projectile

I'm sure you can :-)

> but I really don't get what's the issue with adding some conditional support for Projectile given how simple eglot's usage of project.el is.

That's precisely it: it's simple! Think with me: why should I add a non-essential dependency to eglot when:
1. You tell me that it's not hard to a a `project.el` backend to projectile. (I wouldn't expect it to be, at least not to someone with your skills).
2. There are alternatives in emacs core, where I want Eglot to be headed eventually

[...]

PR #175, opened 2018-12-04
Prefer projectile if available

wyuenho:
In some situations, especially in the case of a monorepo, there will be a different project under different directories, but all under the same repo. LSP servers tend to be located in these subdirectories as well, but Emacs's built-in project package doesn't seem to recognize these different project types (npm, pip etc...) OOTB, so the servers are unable to launch without some .dir-local mess. This PR attempts to fix that.

Aside: The whole point of `project-current` seems to be about getting the root of the current project, removing uses `project-current` will clean up the code somewhat, the downside is `eglot` 's signature will have to change. Let me know if you'd like to change that.

### DIFF
--- a/eglot.el
+++ b/eglot.el
@@ -57,7 +57,8 @@
 
 (require 'json)
 (require 'cl-lib)
-(require 'project)
+(unless (require 'projectile nil t)
+  (require 'project))
 (require 'url-parse)
 (require 'url-util)
 (require 'pcase)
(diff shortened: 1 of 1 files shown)


articuluxe:
Shouldn’t projectile be one implementation of project?  You are free to add more onto project-find-functions.

I don’t use projectile, but another project management system, and would rather see a single interface used...

joaotavora:
A dup of https://github.com/joaotavora/eglot/issues/129, which is marked wontfix.

> Shouldn’t projectile be one implementation of project? You are free to add more onto project-find-functions.

I agree. The author of projectile is presumedly trying to [make this even easier](https://github.com/joaotavora/eglot/issues/129#issuecomment-440284992).

PR #219, opened 2019-02-02
Add aliases for xref-find-references, xref-find-definitions

blue0513:
When I used this package at the first time, I couldn't find how to execute `Go-to-Definition` and `Find-References` because there were no methods like `eglot-go-to-definitions` or so.

IMO, especially for the newbies like me, eglot should allow them to find easily how to execute such functions after typing `M-x eglot-` (then autocompletion will fire and they can find `eglot-find-definitions` & `eglot-find-references`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,12 @@ configuration according to the value of the variable
 `eglot-workspace-configuration`, which you may be set in a
 `.dir-locals` file, for example.
 
+- `M-x eglot-find-definitions` calls `xref-find-definitions` and
+jumps to the definition of the selected symbol.
+
+- `M-x eglot-find-references` calls `xref-find-references` and
+find references of the selected symbol.
+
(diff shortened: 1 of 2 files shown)


blue0513:
Thanks for your quick and polite reply. I really understood what you say!

BTW, this package has been helped me greatly, so I will continue to make PR when I find some bugs and want to improve eglot 👍

dgutov:
I don't really have much to add. The current situation is likely okay (though I don't see a menu item for "find references").

But if you added some aliases, that wouldn't be the end of the world either. It's not like an alias is going to make the maintenance of original commands more difficult.

MaskRay:
The `xref.el` interface should be extended in such a way that makes `textDocument/{declaration,definition,references,typeDefinition,implementation}` possible.

PR #261, opened 2019-05-07
Fix #260: Only consider eglot's own diagnostics in eglot-code-actions

### LINKED ISSUE #260: eglot-code-actions fails when using other Flymake backends
I use my own custom Flymake functions along with eglot, and when I try to do `eglot-code-actions` in a clangd-managed buffer, I get this error:

```

client-request (id:12) Tue May  7 10:59:50 2019:
(:jsonrpc "2.0" :id 12 :method "textDocument/codeAction" :params
	  (:textDocument
	   (:uri "....")
	   :range
	   (:start
	    (:line 0 :character 0)
	    :end
	    (:line 821 :character 0))
	   :context
	   (:diagnostics
	    [(:code "expected_semi_after_expr" :message "Expected ';' after expression (fix available)" :range
		    (:end
		     (:character 79 :line 569)
		     :start
		     (:character 79 :line 569))
		    :severity 1 :source "clang")
	     nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil])))

server-reply (id:12) ERROR Tue May  7 10:59:50 2019:
(:error
 (:code -32600 :message "failed to decode request")
 :id 12 :jsonrpc "2.0")
```

I believe this is because eglot assumes that only it's own Flymake backend is active in a buffer, and it tries to get LSP data from Flymake diagnostics which were not created by eglot.

mkcms:
* eglot.el (eglot-code-actions): Filter out non-eglot diagnostics
  before sending a request to the server.

### DIFF
--- a/eglot.el
+++ b/eglot.el
@@ -2297,10 +2297,9 @@ If SKIP-SIGNATURE, don't try to send textDocument/signatureHelp."
                               :end (eglot--pos-to-lsp-position end))
                  :context
                  `(:diagnostics
-                   [,@(mapcar (lambda (diag)
-                                (cdr (assoc 'eglot-lsp-diag
-                                            (eglot--diag-data diag))))
-                              (flymake-diagnostics beg end))]))))
+                   [,@(cl-loop for diag in (flymake-diagnostics beg end)
+                               when (cdr (assoc 'eglot-lsp-diag (eglot--diag-data diag)))
+                               collect it)]))))
          (menu-items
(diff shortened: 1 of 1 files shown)


mkcms:
> collect it

Whoah, I didn't know that this `it` keyword was available in `cl-loop`, pretty cool. I pushed your version.

PR #349, opened 2019-11-16
Remove the display of (snippet) in completion annotation

xuchunyang:
I'm using Eglot with Gopls. During code completion, `(snippet)` shows all the time, even I'm completing a variable such as `os.O_RDONLY`, I'm not sure it's a bug of Eglot. Anyway it is not useful for me and takes some valuable space, so I want it to be stripped.

Screen Shot 2019-11-16 at 20 16 47

### DIFF
--- a/eglot.el
+++ b/eglot.el
@@ -2035,10 +2035,7 @@ is not active."
              (when annotation
                (concat " "
                        (propertize annotation
-                                   'face 'font-lock-function-name-face)
-                       (and (eql insertTextFormat 2)
-                            (eglot--snippet-expansion-fn)
-                            " (snippet)"))))))
+                                   'face 'font-lock-function-name-face))))))
        :company-doc-buffer
        (lambda (proxy)
          (let* ((documentation


nemethf:
I find that information useful when I use other serves that send snippets and normal completion items at the same time.  If there's not enough space to display the whole text, then I think it's a better approach to let the UI trim the right part of the text. 

(Additionally, we received a few issues when users failed to configure snippet completion.  It is a good visual feetback of a successful configuration.)

xuchunyang:
OK, that isn't very surprising, there must be a reason for it. However, I've not find it useful and it's a bit of annoying because my screen is small and I also have to use large font for my poor eyesight.

> then I think it's a better approach to let the UI trim the right part of the text.

Unfortunately, company-mode doesn't trim long line, and it looks very ugly. I guess it's a known issue of company-mode for a long time and unlikely to be fixed any time soon.

Screen Shot 2019-11-17 at 01 38 13

I think I'm just going to edit the code to apply the change, it's not ideal but can work.

nemethf:
@xuchunyang, I formed just another user option and did not write with "authority".  If I convinced you, that's great, but I think it's usually a good idea to wait for the project leader to close these kinds of pull requests.  

>  I guess it's a known issue of company-mode for a long time

I'm not so sure about that.

joaotavora:
> OK,  that isn't very surprising, there must be a reason for it. 

Actually there isn't much reason. I needed that when I was developing Eglot, but I doubt it's much use.

I was gonna just merge your PR and forget about it. But I'm going to analyze this better after Felician expressed his opinion...

joaotavora:
> Additionally, we received a few issues when users failed to configure snippet completion. It is a good visual feetback of a successful configuration.)

This is true, but the "snippet" is so annoying that I don't think it's worth it. And those reports have died down a bit lately, maybe due to the stuff added to the README.

So I'm leaning towards merging this. An alternative would be to check if a particularly useful utf-8 char is displayable and use that instead.

joaotavora:
> I'm not so sure about that.

Yeah, that second company screenshot looks like a bug, you might want to report it, @xuchunyang

joaotavora:
@nemethf I merged this, but as I said, we can try to do something like

```lisp
(if (and (eql insertTextFormat 2)
         (eglot--snippet-expansion-fn)
         (char-displayable-p ?\ⓢ))
    "ⓢ"
  " ")
```

nemethf:
I now think nobody will select a completion item based on its "snippetness", therefore the information is pointless under normal usage.  (But it might be useful for server developers, for example.)   The character "ⓢ" is nice, I think it is too cryptic for newbies.  So I'm OK with this change.  Sorry for the noise.

PR #408, opened 2020-01-11
Support markdown only when gfm-view-mode is installed

nemethf:


theothornhill:
Maybe you can add that with `(setq markdown-fontify-code-blocks-natively t)` code blocks inside the `gfm-view` will be fontified. IMO this should be default behaviour in eglot, but I guess it isn't eglots job to tweak other packages.

joaotavora:
> (setq markdown-fontify-code-blocks-natively t)

Why isn't this the default behaviour of `gfm-view` itself?

README.md
```diff
@@ -414,6 +414,10 @@ queries the server for possible code actions at point.  See variable
 ## Hover on symbol
 ![eglot-hover-on-symbol](./gif-examples/eglot-hover-on-symbol.gif)
 
+It is not shown above, but if [markdown-mode][markdown] is installed,
```
joaotavora, line 417: Only read this now.  Really it would be ideal to make another gif instead of saying this, don't you think.  Can you give me a good setup for recording one?  Maybe python? I haven't been using eglot a lot lately.

theothornhill:
> Why isn't this the default behaviour of gfm-view itself?

I have no idea... I can ask them, but meanwhile, it may be a good idea to include in eglot docs. It is a bit hard to find otherwise, don't you think?

joaotavora:
> I have no idea... 

We should ask them

> I can ask them, but meanwhile, it may be a good idea to include in eglot docs. It is a bit hard to find otherwise, don't you think?

If we find that to be a suitable default, it wouldn't be setting a precedent to let-bind it or locally set it that temporary buffer. We do the same for `company-backends` and `company-tooltip-annotations`.  But it would be nice to know what it does exactly first.

joaotavora:
> I have no idea... 

We should ask them

> I can ask them, but meanwhile, it may be a good idea to include in eglot docs. It is a bit hard to find otherwise, don't you think?

If we find that to be a suitable default, it wouldn't be setting a precedent to let-bind it or locally set it that temporary buffer. We do the same for `company-backends` and `company-tooltip-annotations`.  But it would be nice to know what it does exactly first.

theothornhill:
> We should ask them 

See [issue](https://github.com/jrblevin/markdown-mode/issues/451)

This is without the option:
Screenshot from 2020-04-28 22-43-25

This is with the option:
Screenshot from 2020-04-28 22-41-52

> If we find that to be a suitable default, it wouldn't be setting a precedent to let-bind it or locally set it that temporary buffer. 

`(setq-local markdown-fontify-code-blocks-natively t)` should work fine. It also needs to set a default variable, `(setq markdown-fontify-code-block-default-mode major-mode)` if it cannot derive from the code block itself.

README.md
nemethf: Sure.  I don't use gfm-view myself, but I can install it and record an animgif for this.  

I know disk space is cheap, but it strangely annoys me when an otherwise small git repository is full of large gif.  Would it be possible to store the new ones elsewhere?  Like on the wiki, or a separate repository.

nemethf:
> `(setq-local markdown-fontify-code-blocks-natively t)` should work fine. It also needs to set a default variable, `(setq markdown-fontify-code-block-default-mode major-mode)` if it cannot derive from the code block itself.

I think if it cannot derive from the code block, then Eglot shouldn't try to be smart here. Do you have an example where the server doesn't indicate the language at the heading of a code block?

theothornhill:
> I think if it cannot derive from the code block, then Eglot shouldn't try to be smart here. Do you have an example where the server doesn't indicate the language at the heading of a code block?

I agree, and no, I haven't seen code blocks without annotation yet, but I'm sure it exists somewhere. If I understand correctly, it looks like it will only default back to how it looks right now if it cannot find anything, and I guess that is fine.

theothornhill:
gfm-view-mode now fontifies natively - they were quick to incorporate it. Should work out of the box now in eglot.

Issue #440, opened 2020-04-23
Improving eglot-eldoc-function


theothornhill:
Hello!

I've tried to simplify a little the view we get when using the normal onHover. The previous functionality was to just display the whole response in the minibuffer. This creates a lot of jumping around with large buffers. I know there are workarounds for this, like the `max-mini-window-height`, but still, I guess as a good emacs citizen we shouldn't really use more than one line?

To accomodate for this I've created a new defcustom, so that it is an opt-in feature. When it is set to `t`, we get only the `:value` from the markup received. 

There is one issue with this first hand-wavy implementation - the MarkupContent response. It doesn't really fit into the whole display in minibuffer concept, and usually returns a huge string. 

Right now the `eglot-help-at-point` doesn't pass the new arg, and therefore we still get the whole response when we want. 

Could this go forward? I believe we need some better logic in the `eglot--hover-info` to handle the different responses, but right now it works fine in `rls`, `elm-language-server` and `fsac`(fsharp)

Have a nice day!
```diff
diff --git a/eglot.el b/eglot.el
index 13ee307..f8da931 100644
--- a/eglot.el
+++ b/eglot.el
@@ -1106,7 +1106,7 @@ Doubles as an indicator of snippet support."
   (pcase-let ((`(,string ,mode)
                (if (stringp markup) (list (string-trim markup)
                                           (intern "gfm-view-mode"))
-                 (list (plist-get markup :value)
+                 (list (string-trim (plist-get markup :value))
                        (pcase (plist-get markup :kind)
                          ("markdown" 'gfm-view-mode)
                          (_ major-mode))))))
@@ -2131,11 +2131,13 @@ is not active."
 
 (defvar eglot--highlights nil "Overlays for textDocument/documentHighlight.")
 
-(defun eglot--hover-info (contents &optional range)
+(defun eglot--hover-info (contents &optional range show-only-signature)
   (let ((heading (and range (pcase-let ((`(,beg . ,end) (eglot--range-region range)))
                               (concat (buffer-substring beg end)  ": "))))
-        (body (mapconcat #'eglot--format-markup
-                         (if (vectorp contents) contents (list contents)) "\n")))
+        (body (if (and show-only-signature (null (plist-get contents :kind)))
+                  (eglot--format-markup (aref contents 0))
+                  (mapconcat #'eglot--format-markup
+                             (if (vectorp contents) contents (list contents)) "\n"))))
     (when (or heading (cl-plusp (length body))) (concat heading body))))
[...]

PR #443, opened 2020-04-23
Respect eldoc-echo-area-use-multiline-p

muffinmad:
Make `eglot-eldoc-function` handle the variable `eldoc-echo-area-use-multiline-p`.

### DIFF
--- a/eglot.el
+++ b/eglot.el
@@ -2247,36 +2247,48 @@ Buffer is displayed with `display-buffer', which obeys
   "Put updated documentation STRING where it belongs.
 Honours `eglot-put-doc-in-help-buffer'.  HINT is used to
 potentially rename EGLOT's help buffer."
-  (if (or (eq t eglot-put-doc-in-help-buffer)
-          (and eglot-put-doc-in-help-buffer
-               (funcall eglot-put-doc-in-help-buffer string)))
-      (with-current-buffer (eglot--help-buffer)
-        (rename-buffer (format "*eglot-help for %s*" hint))
-        (let ((inhibit-read-only t))
-          (erase-buffer)
-          (insert string)
-          (goto-char (point-min))
-          (if eglot-auto-display-help-buffer
-              (display-buffer (current-buffer))
-            (unless (get-buffer-window (current-buffer))
-              (eglot--message
-               "%s\n(...truncated. Full help is in `%s')"
-               (truncate-string-to-width
-                (replace-regexp-in-string "\\(.*\\)\n.*" "\\1" string)
-                (frame-width) nil nil "...")
-               (buffer-name eglot--help-buffer))))
-          (help-mode)))
-    (eldoc-message (and string
-                        (if (eq eldoc-echo-area-use-multiline-p t)
-                            string
-                          (let ((string (replace-regexp-in-string
-                                        "\\(.*\\)\n.*" "\\1" string))
-                                (ea-width (1- (window-width (minibuffer-window)))))
-                            (if (null eldoc-echo-area-use-multiline-p)
-                                (substring string 0 ea-width)
-                              ;; TODO If showing signatures and string doesn't fit,
-                              ;; strip function name and leave only args.
-                              string)))))))
+  (when (or (eq t eglot-put-doc-in-help-buffer)
+            (and eglot-put-doc-in-help-buffer
+                 (funcall eglot-put-doc-in-help-buffer string)))
+    (with-current-buffer (eglot--help-buffer)
+      (rename-buffer (format "*eglot-help for %s*" hint))
+      (let ((inhibit-read-only t))
+        (erase-buffer)
+        (insert string))
+      (goto-char (point-min))
+      (help-mode)
+      (when eglot-auto-display-help-buffer
+        (display-buffer (current-buffer)))))
+  (unless (and eglot--help-buffer
+               (get-buffer-window eglot--help-buffer))
+    (eldoc-message
+     (let* ((info-lines (split-string string "\n"))
+            (signature-line (car info-lines)))
+       ;; Special interpretation of `eldoc-echo-area-use-multiline-p':
+       ;; If t, show additional info lines.  Show only signature line
+       ;; otherwise.
+       (if (eq eldoc-echo-area-use-multiline-p t)
+           ;; limit message to mini-window height
+           (let* ((info-lines (cdr info-lines))
+                  (ea-width (window-width (minibuffer-window)))
+                  (lines-avail
+                   (- (if (floatp max-mini-window-height)
+                          (floor (* max-mini-window-height (frame-height)))
+                        max-mini-window-height)
+                      (ceiling (length signature-line) ea-width)))
+                  (lines (list signature-line)))
+             (catch 'done
+               (dolist (line info-lines)
+                 (when (<= lines-avail 0)
+                   (throw 'done t))
+                 (let ((add-line (truncate-string-to-width
+                                  line (* lines-avail ea-width) nil nil "...")))
+                   (push add-line lines)
+                   (setq lines-avail (- lines-avail
+                                        (ceiling (max 1 (length add-line)) ea-width))))))
+             (string-join (nreverse lines) "\n"))
+         ;; TODO if signature line doesn't fit, strip function name
+         signature-line)))))
 
 (defun eglot-eldoc-function ()
   "EGLOT's `eldoc-documentation-function' function."
--- a/eglot.el
+++ b/eglot.el
@@ -2248,6 +2248,46 @@ Buffer is displayed with `display-buffer', which obeys
 `display-buffer-alist' & friends."
   :type 'boolean)
 
+(defcustom eglot-documentation-function 'eglot-documentation-singleline
+  "Function to prepare hover and signature info for display.
+Function is called with documentation string as argument.
+It's return value will be shown via `eldoc-message'."
+  :type '(choice (const :tag "Display first line" eglot-documentation-singleline)
+                 (const :tag "Display multiple lines" eglot-documentation-multiline)
+                 function))
+
+(defun eglot-documentation-singleline (string)
+  "Return first line from STRING.
+Line can be truncated with respect to `eldoc-echo-area-use-multiline-p'."
+  (let ((line (car (split-string string "\n"))))
+    (if eldoc-echo-area-use-multiline-p
+        ;; TODO strip function name if `eldoc-echo-area-use-multiline-p' is
+        ;; 'truncate-sym-name-if-fit and signature line doesn't fit.
+        line
+      (truncate-string-to-width
+       line (window-width (minibuffer-window)) nil nil "..."))))
+
+(defun eglot-documentation-multiline (string)
+  "Return as much text from STRING as echo area can display."
+  (let* ((info-lines (split-string string "\n"))
+         (ea-width (window-width (minibuffer-window)))
+         (lines-avail
+          (if (floatp max-mini-window-height)
+                 (floor (* max-mini-window-height (frame-height)))
+               max-mini-window-height))
+         lines)
+    (catch 'done
+      (dolist (line info-lines)
+        (when (<= lines-avail 0)
+          (throw 'done t))
+        (let ((add-line (truncate-string-to-width
+                         line (* lines-avail ea-width) nil nil "...")))
+          ;; Maybe we can skip empty lines to fit more text
+          (push add-line lines)
+          (setq lines-avail (- lines-avail
+                               (ceiling (max 1 (length add-line)) ea-width))))))
+    (string-join (nreverse lines) "\n")))
+
 (defun eglot--update-doc (string hint)
   "Put updated documentation STRING where it belongs.
 Honours `eglot-put-doc-in-help-buffer'.  HINT is used to
@@ -2269,33 +2309,7 @@ potentially rename EGLOT's help buffer."
   (unless (and eglot--help-buffer
                (get-buffer-window eglot--help-buffer))
     (eldoc-message
-     (let* ((info-lines (split-string string "\n"))
-            (signature-line (car info-lines)))
-       ;; Special interpretation of `eldoc-echo-area-use-multiline-p':
-       ;; If t, show additional info lines.  Show only signature line
-       ;; otherwise.
-       (if (eq eldoc-echo-area-use-multiline-p t)
-           ;; limit message to mini-window height
-           (let* ((info-lines (cdr info-lines))
-                  (ea-width (window-width (minibuffer-window)))
-                  (lines-avail
-                   (- (if (floatp max-mini-window-height)
-                          (floor (* max-mini-window-height (frame-height)))
-                        max-mini-window-height)
-                      (ceiling (length signature-line) ea-width)))
-                  (lines (list signature-line)))
-             (catch 'done
-               (dolist (line info-lines)
-                 (when (<= lines-avail 0)
-                   (throw 'done t))
-                 (let ((add-line (truncate-string-to-width
-                                  line (* lines-avail ea-width) nil nil "...")))
-                   (push add-line lines)
-                   (setq lines-avail (- lines-avail
-                                        (ceiling (max 1 (length add-line)) ea-width))))))
-             (string-join (nreverse lines) "\n"))
-         ;; TODO if signature line doesn't fit, strip function name
-         signature-line)))))
+     (funcall eglot-documentation-function string))))
 
 (defun eglot-eldoc-function ()
   "EGLOT's `eldoc-documentation-function' function."
--- a/eglot.el
+++ b/eglot.el
@@ -2266,7 +2266,17 @@ potentially rename EGLOT's help buffer."
                 (frame-width) nil nil "...")
                (buffer-name eglot--help-buffer))))
           (help-mode)))
-    (eldoc-message string)))
+    (eldoc-message (and string
+                        (if (eq eldoc-echo-area-use-multiline-p t)
+                            string
+                          (let ((string (replace-regexp-in-string
+                                        "\\(.*\\)\n.*" "\\1" string))
+                                (ea-width (1- (window-width (minibuffer-window)))))
+                            (if (null eldoc-echo-area-use-multiline-p)
+                                (substring string 0 ea-width)
+                              ;; TODO If showing signatures and string doesn't fit,
+                              ;; strip function name and leave only args.
+                              string)))))))
 
 (defun eglot-eldoc-function ()
   "EGLOT's `eldoc-documentation-function' function."


joaotavora:
Shouldn't `eldoc-message` be doing this?  I.e. be honouring `eldoc-echo-are-use-multiline-p`?

@monnier can we have your input here? I know `eglot` shouldn't call `eldoc-message` directly, but it has few other options given the asynchronous nature of LSP.  

@muffinmad, I think you could implement this feature with an `:around` to `eldoc-message`.  I don't like seeing it there. 

But the idea in general seems correct.

muffinmad:
> Shouldn't `eldoc-message` be doing this? I.e. be honouring `eldoc-echo-are-use-multiline-p`?

Well, docstring of the `eldoc-documentation-function` notes that the function (`eglot-eldoc-function` in our case) must explicitly handle `eldoc-echo-are-use-multiline-p`.

Just wanted eldoc to show first line from hover info. Was going to create option like `eglot-eldoc-show-first-line` or something and then found `eldoc-echo-are-use-multiline-p`.
 
Did you notice how symbol name is stripped in eldoc message while editing elisp? E.g.:
1. emacs -Q
2. eval `(set-frame-width nil 62)`
3. Type `(replace-regexp-in-string` and wait for eldoc message

Echo area will contain only args. Can we have this feature in eglot, please? :)

> @muffinmad, I think you could implement this feature with an `:around` to `eldoc-message`. I don't like seeing it there.

Sure, no problem. Should I close this PR?

joaotavora:
> Sure, no problem. Should I close this PR?

No, let me think about this. And wait for Stefan's opinion.

monnier:
> Shouldn't `eldoc-message` be doing this? I.e. be honouring `eldoc-echo-are-use-multiline-p`?

To the extent that it's kinda meant to be internal, there is no
right/wrong answer to this question.

> @monnier can we have your input here? I know `eglot` shouldn't call
> `eldoc-message` directly, but it has few other options given the
> asynchronous nature of LSP.

I think that's the problem that needs to be fixed first: we need to
provide some official way for an eldoc backend to work asynchronously.

It shouldn't be hard to change `eldoc.el` to provide that.  [ Once that's
done, we can see how to solve the problem of Eglot needing to use it and
to also work on older Emacsen.  ]

joaotavora:
> I think that's the problem that needs to be fixed first: we need to
provide some official way for an eldoc backend to work asynchronously.

Any ideas? The best I can see is to replicate `flymake.el`'s approach here.  `eldoc-documentation-function` would be passed a callbak to be funcalled when the backend deems the doc ready to present.  I can't see it working with multiple backends though (`add-function`-style multiple backends), but perhaps it's not important.  The greatest drawback is that this passing can't be via arglist, so we would have to use some other way.  Special var?  And return a non-nil, non-string to indicate that `eldoc` should assume the backend is going to call the function eventually.

>  [ Once that's
done, we can see how to solve the problem of Eglot needing to use it and
to also work on older Emacsen.  ]

As always, I'd like to revert that. First the `:core` thingy, then the fix.  It makes developing on Eglot much easier mainly because I don't force my collaborators to use bleeding edge Emacsen.

You know what I'm going to suggest here right?

joaotavora:
> As always, I'd like to revert that.

Doh.  I meant "reverse".

joaotavora:
> You know what I'm going to suggest here right? 

Doh x2.  This sentence doesn't make sense. Sorry.

nemethf:
It seems the discussion is shifted towards eldoc in general, so I'd like to mention that the only problem I have with multiline eldoc is that it shows the last part of a long message.  I found the first part of a long documentation was more useful than the last part was. 

Eglot can put long documentation in eglot--help-buffer, which is even more useful because I can C-M-v and C-M-S-v to read it in its entirety.  However, the documentation in eldoc is automatically appears and disappears and this feature is missing when eglot-put-doc-in-help-buffer is t.  And it seems the C-M-v feature doesn't work.  (I'm going to open an issue about it.)

theothornhill:
The above referenced issue #440 deals with the MarkupContent not being shown, and addresses part of @nemerthf concern.

joaotavora:
> Eglot can put long documentation in eglot--help-buffer, which is even more useful because I can C-M-v and C-M-S-v to read it in its entirety. However, the documentation in eldoc is automatically appears and disappears and this feature is missing when eglot-put-doc-in-help-buffer is t. And it seems the C-M-v feature doesn't work. (I'm going to open an issue about it.)

Yes, and that issue _should_ have a an easy fix.  I just answered in #445.

Meanwhile, @muffinmad I think we should merge your PR, since it's localized enough and not too complicated.  However, it seem like `pcase` or `case` or even `cond` could simplify it, I'm not a fan of those two `if`s.

muffinmad:
> Meanwhile, @muffinmad I think we should merge your PR, since it's localized enough and not too complicated. However, it seem like `pcase` or `case` or even `cond` could simplify it, I'm not a fan of those two `if`s.

@joaotavora Me either. But first `if` can work without calling `replace-regexp-in-string` and `window-width`. However I think first `if` condition will be `t` rarely,  I will update PR to use `cond`.

joaotavora:
> However I think first if condition will be t rarely, I will update PR to use cond.

Wait, don't waste the effort yet.  I need to think about this.  There is some overlap `eglot-put-doc-in-help-buffer`, which also allows functions that tell if the remainder of the doc should be placed in the help buffer and only the first line shown.  Help me understand your case, what do you have that variable set to?

muffinmad:
My goal is to display first line of hover info in echo area. I can't achieve this by settings `eglot-put-doc-in-help-buffer` to `t`:
1. String is truncated to frame width. Last arguments in signature info are not visible. 
2. `[eglot] ` prefix of `eglot-message` is not taken into account. String is wrapped to second line.
3. Next line in echo area is occupied by `(...truncated...` message. Even if string wasn't truncated.
4. `eglot-message` is hidden on every keypress while `eldoc-message` persist.

Issues 2 and 3 can be solved easily. But 1 and 4 are game changers. Especially 4: having persistent signature info in echo area while eglot is highlighting current arg is awesome.

Now with the changes from this PR, I set the `eglot-put-doc-in-help-buffer` to `nil`.

You are right indeed about two stripping methods overlap in one function. And I for one prefer `eldoc-message` over `eglot--message`.

theothornhill:
> My goal is to display first line of hover info in echo area. I can't achieve this by settings `eglot-put-doc-in-help-buffer` to `t`:

#440 Deals with exactly this. As far as I can tell there are three interfaces allowed for the hover response, one which doesn't really work, since it's a huge string and there is no way to find what is the signature.

The other responses however, leaves the

 `(:language "something" :value "signature here")`

Which can be accessed. I've added a new defcustom to deal with this, so the `eglot--hover-info` can be used both in `eglot-help-at-point` and on normal hover.

It seems like it fits in this pr, and thats why I've brought it up, or am I completely missing the point?

joaotavora:
> It seems like it fits in this pr, and thats why I've brought it up, or am I completely missing the point?

You're not.  I think you're more or less solving the same problem yes.  The difference and the reason i'm more slightly more receptive to @muffinmad 's version is that it uses an existing variable in Emacs, which is a always a good thing (slightly less Eglot complexity).  That said, it's not complete: it doesn't deal with non-t, non-nil values that `eldoc-echo-are-use-multiline-p` can have.  I'll try to complete it afterwards: i'm mostly convinced it can coexist with `eglot-put-in-help-buffer`.

The intervening discussion about an `eldoc.el` redesign, if it were to bear fruit, could probably deal with a lot of this.  But that's not a goal for the short-term.

monnier:
>> I think that's the problem that needs to be fixed first: we need to
>> provide some official way for an eldoc backend to work asynchronously.
> Any ideas? The best I can see is to replicate `flymake.el`'s approach here.

Nothing specific, no, I haven't looked at the problem in enough detail.

> `eldoc-documentation-function` would be passed a callbak to be funcalled

[ Note: The current eldoc.el now uses `eldoc-documentation-function*s*`.  ]

That's one option, indeed.

Here's another:

When working asynchronously, the function can return nil immediately
and call some new `eldoc-documentation-callback` function later, when it
gets its answer.

The package can also register on a new `eldoc-documentation-abort-hook`
which is called whenever we should abort any potential async request
that's still in progress.

>> [ Once that's done, we can see how to solve the problem of Eglot
>>   needing to use it and to also work on older Emacsen.  ]
> As always, I'd like to reverse that.  First the `:core` thingy, then the fix.

One of the possible fixes *is* "the :core thing".


        Stefan

muffinmad:
> As far as I can tell there are three interfaces allowed for the hover response, one which doesn't really work, since it's a huge string and there is no way to find what is the signature.

Let's leave it to server to send signature in first line of that huge string in onHover response.

> The other responses however, leaves the
>
> `(:language "something" :value "signature here")`

This (deprecated) case is properly handled by `eglot--sig-info`. E.g. this `MarkedString[]`:

```
[(:value "foo(bar)" :language "python")
	    "docstring for foo"]
```

will become this string:

```
foo(bar)
docstring for foo
```

Again, signature is on the first line.

> Which can be accessed. I've added a new defcustom to deal with this, so the eglot--hover-info can be used both in eglot-help-at-point and on normal hover.

IMO there are no need in `eglot-eldoc-show-only-signature` option. It must always be `t`.

Let's take a look at elisp: only function signature is displayed in the echo area. To get docstring one must use `describe-function`.

I think Eglot should work the same way: always display first line of hover info in the echo area. To get full hover info user can use `eglot-help-at-point`.

If I were allowed, I would delete `eglot-put-doc-in-help-buffer` and `eglot-auto-display-help-buffer` options. However, if someone want to display full hover info as point moves, we can provide some kind of hook for it. `eglot-update-doc-hook` or something. That hook will be called with full hover info as the argument. And that hook can do whatever he wants with full hover info: show it in separate window/frame, write it to log, pass it to text-to-speech service, etc.

At least let's remove displaying hover/signature info with `eglot--message`.

joaotavora:
@monnier:

> When working asynchronously, the function can return nil immediately
> and call some new `eldoc-documentation-callback` function later, when it
> gets its answer.

OK, but is that a variable? Set by whom and holding what? A closure? A top-level anonymous function or a named function?

> One of the possible fixes *is* "the :core thing".

Is upgrading to the latest `eldoc.el` a fix in itself? That's great, but I don't  understand how.

monnier:
>> When working asynchronously, the function can return nil immediately
>> and call some new `eldoc-documentation-callback` function later, when it
>> gets its answer.
> OK, but is that a variable? Set by whom and holding what? A closure?
> A top-level anonymous function or a named function?

I was thinking of a top-level function named `eldoc-documentation-callback`.

>> One of the possible fixes *is* "the :core thing".
> Is upgrading to the latest `eldoc.el` a fix in itself? That's great,
> but I don't understand how.

No, I just meant that after we decide what the future solution will look
like within eldoc.el, we can start thinking about how to let packages
work with older Emacsen, either by requiring the new version of eldoc.el
(and distributing it via :core), or by providing a forward-compatibility
eldoc-async.el package in GNU ELPA, or by making it possible/easy to
degrade gracefully, ...

The important aspect is that those are 2 separate steps and we shouldn't
care about the second step when designing the first.


        Stefan

joaotavora:
> The important aspect is that those are 2 separate steps and we shouldn't
care about the second step when designing the first.

Sure, I agree.  But I also think I don't want to spend much time designing a thing now that will only be available for Emacs 28.  So if it becomes available for, say, >= 26.3 I'll spend more time on it.  So I'd rather get that over with first and then move on to the design.

Speaking of which I don't understand your idea. `eldoc-documentation-callback` is called when the backend sees fit (when it has received the docstring, probably from a process filter).  And what does the backend give the function? What does the function do? Why is it called "callback" and how does this differ from the current way the backend calls `eldoc-message`?

monnier:
> Sure, I agree.  But I also think I don't want to spend much time designing
> a thing now that will only be available for Emacs 28.  So if it becomes
> available for, say, >= 26.3 I'll spend more time on it.

I can't think of any reason why it couldn't be available for 26.3, so
I wouldn't worry about that part.


        Stefan

monnier:
> Speaking of which I don't understand your idea.
> `eldoc-documentation-callback` is called when the backend sees fit
> (when it has received the docstring, probably from a process filter).

We'd have to define when it's allowed to call it, but ... yes, pretty much.

> And what does the backend give the function?

We still have to define that, but I guess it would pass it the value
that it would have otherwise returned synchronously, maybe together with
some info (such as some kind of backend identifier?) so eldoc.el can
keep the info from various backends properly ordered?

> What does the function do?

Whatever is usually done when the synchronous backend call returns a string?

> Why is it called "callback" and how does this differ from the current
> way the backend calls `eldoc-message`?

`eldoc-message` was not designed to be called as a callback.  It may
work well enough most of the time, but `eldoc-documentation-callback`
documented to be usable for the asynchronous case, and hence should be
be more robust.


        Stefan

nemethf:
> IMO there are no need in eglot-eldoc-show-only-signature option. It must always be t.
>
> Let's take a look at elisp: only function signature is displayed in the echo area. To get docstring one must use describe-function.
>
> I think Eglot should work the same way: always display first line of hover info in the echo area. To get full hover info user can use `eglot-help-at-point`.

I kind of disagree.  I find valuable to instantly see the whole info without pressing any keys to see or hide it.  

The only problem I had was that for a 10 line long documentation it displayed the last 6 and not the first 6 lines.  I wrote [something](https://github.com/nemethf/bess-language-server/blob/develop/bess.el#L14) to address this, which works, but I haven't proposed for inclusion because I never got around to cleaning and simplifying it.

muffinmad:
@nemethf Please see updated PR. With `eldoc-echo-area-use-multiline-p` set to `t` additional top lines from hover info will be shown in the echo area. With `eldoc-echo-area-use-multiline-p` set to any other value only first line of hover info will be shown.

muffinmad:
@theothornhill Could you also check if updated PR works for you?

theothornhill:
> Could you also check if updated PR works for you?

I'd say it kinda works. Not sure if you have rust, but this was the
simplest way for me to check. 

    use std::io;
    
    fn main() {
    
        println!("Please input your guess.");
    
        let mut guess = String::new();
    
        io::stdin()
            .read_line(&mut guess)
            .expect("Failed to read line");
    
        println!("You guessed: {}", guess)
    
    }

Given this snippet, I get a few bugs:

- When hovering `guess`, it shows how I want it in the
  minibuffer. However, it doesn't go away if I later hover `String` and
  move again. Same for `main`.

- Also I would not want to see the "truncated" message, since, in my
  opinion it isn't really relevant to the minibuffer.

It is still an improvement though :)

theothornhill:
> Could you also check if updated PR works for you?

I'd say it kinda works. Not sure if you have rust, but this was the
simplest way for me to check.

```rust
use std::io;

fn main() {
    let mut guess = String::new();

    io::stdin()
        .read_line(&mut guess)
        .expect("Failed to read line");
}
```
Given this snippet, I get a few bugs:

- When hovering `guess`, it shows how I want it in the
  minibuffer. However, it doesn't go away if I later hover `String` and
  move again. Same for `main`.

- Also I would not want to see the "truncated" message, since, in my
  opinion it isn't really relevant to the minibuffer.

It is still an improvement though :)

muffinmad:
> Not sure if you have rust

I don't have it. So instead of me starting journey in rust world can you please answer few questions :)

> * When hovering `guess`, it shows how I want it in the
>   minibuffer. However, it doesn't go away if I later hover `String` and
>   move again. Same for `main`.

Did server return some hover info for `String`? If not, then it's #439.

> * Also I would not want to see the "truncated" message, since, in my
>   opinion it isn't really relevant to the minibuffer.

What truncated message are you talking about? What do you have `eldoc-echo-area-use-multiline-p` set to?

theothornhill:
> What truncated message are you talking about? What do you have eldoc-echo-area-use-multiline-p set to?

`truncate-sym-name-if-fit` - I think it is the default?

> Did server return some hover info for String?

`(:language "rust" :value "pub struct String")`, so I don't think it is #439 

And it looks like this:
Screenshot from 2020-04-27 18-40-06

When  hovering `guess` I get this:
Screenshot from 2020-04-27 18-41-36


If I now move back over to `String` it shows like the first picture again, but if i then move one line up, it still shows `guess` without me hovering that. In normal eldoc behaviour it is cleared after a second or something.

Like this:
Screenshot from 2020-04-27 18-43-42

To me it seems like when I hover `guess` it gets stored, then when I hover `String` it is not overridden, and `String` isn't stored.

I know there probably are several things here, some maybe not even related at all to your PR, but I wanted to point it out anyways.

joaotavora:
(@theothornhill this is off-topic but that is an exquisitely beautiful theme)

theothornhill:
> (@theothornhill this is off-topic but that is an exquisitely beautiful theme)

Hehe, thanks! It is [gruvbox-dark-hard](https://github.com/Greduan/emacs-theme-gruvbox), and [mood-line](https://gitlab.com/jessieh/mood-line/-/tree/master)

muffinmad:
> `truncate-sym-name-if-fit` - I think it is the default?

Yes it is.

> And it looks like this:
> Screenshot from 2020-04-27 18-40-06

Well, I don't know where this "truncated..." message come from - this PR removes it from `eglot.el`.

> 
> When hovering `guess` I get this:
> Screenshot from 2020-04-27 18-41-36

Isn't hover correct here? Looks like it show that `guess` is `std::string::String`.

> Like this:
> Screenshot from 2020-04-27 18-43-42

And this is definitely #439.

theothornhill:
> Well, I don't know where this "truncated..." message come from - this PR removes it from eglot.el.

My bad. I cloned your repo and evaluated it on master. I assumed your commit `merge master` merged your changes into your master for some reason. When I checked out `eldoc-multiline` it definitively works now. Good job, and sorry for the confusion...

Good job :)

joaotavora:
> Well, I don't know where this "truncated..." message come from - this PR removes it from eglot.el

It comes from Eglot.  We shouldn't remove it from `eglot.el` entirely, though it seems to be true that we should remove it from this particular case.

> Isn't hover correct here? Looks like it show that guess is std::string::String.

Yes, there's a little ambiguity in `trucate-sym-name-if-fit`.  It says "_symbol name may be truncated if it will enable the function arglist or documentation string to fit on a single line without resizing window_" This seems to state that only if that truncating action happens to fix the one-line problem should we do it.  I think it is too complicated.  Don't know what to do here.

nemethf:
> @nemethf Please see updated PR. With `eldoc-echo-area-use-multiline-p` set to `t` additional top lines from hover info will be shown in the echo area. With `eldoc-echo-area-use-multiline-p` set to any other value only first line of hover info will be shown.

I think that's not really in line with the documentation of `eldoc-echo-area-use-multiline-p`.   If this variable is set to `truncate-sym-name-if-fit`, then (if I understand the documentation correctly) eldoc can remove only the name of the symbol.

If the point is on "sys" in a python buffer, then I get the following one line in the echo area:
"This module provides access to some objects used or maintained by the".
The hover text is so large that it surely needs to truncated, but it's surprising that only the first line is shown because my `eldoc-echo-area-use-multiline-p` is `truncate-sym-name-if-fit`.

So, instead of checking whether eldoc-echo-area-use-multiline-p is t in eglot--update-doc, I think it would be a bit better to check whether it is non-nil.  However, I understand I might be the only one here who wants to see the full hover info, and I can set this variable to t.  Additionally, your PR makes eglot/eldoc more useful even in my use-case, so I welcome it either way.  Thanks.

muffinmad:
Hello everybody!

In updated version of PR (which see) you may find new user option `eglot-documentation-function`. There are also two predefined functions to set as value of the `eglot-documentation-function`: `eglot-documentation-singleline` and `eglot-documentation-multiline`.
Those of us, who prefer compact yet informative hover info, like me and @theothornhill, will have what they want OOTB.
Those of us, who prefer full hover info blasting their echo area, like @nemethf, may set `eglot-documentation-function` to `eglot-documentation-multiline`.

Thanks for your attention!

joaotavora:
> In updated version of PR (which see) you may find new user option eglot-documentation-function.

I'm sorry I haven't had much time to look into this, but I thought this PR was a means to respect `eldoc-echo-are-use-multiline-p`, very much as its title states.  Do we all agree that we need a new user options to do this, i.e. to respect this variable?  That would make it 3 variables to control documentation:

- `eglot-put-doc-in-help-buffer`
- `eldoc-echo-area-use-multiline-p`
- the proposed new variable, whose name is easy to mix with `eldoc-documentation-function` and the existing value for it, `eglot-eldoc-function`.

I'm not sure I'm too fond of this idea...

joaotavora:
So, I've "closed" this is a commit, kind of tongue-in-cheek.  I think i pushed the lowest common denominator for this, certainly the simplest I think.  Maybe it doesn't do everything you want. I wonder what each of you thinks of this approach, its drawbacks, etc.

theothornhill:
I could not get "my" use-case to work properly, where when I hover, eglot only shows first line, and if I `M-x eglot-help-at-point`, then the whole thing appears.

When I fiddled a bit with the code to check why it wasn't working, I noticed both the truncate part and "my" use-case uses same-ish logic, so why couldn't they use the same?

This diff shows a small proposal which I think makes the code a bit easier to read, and still respects the `cond` you created. In addition, to get the behaviour I want, I have to adjust eldoc variables accordingly

```diff
diff --git a/eglot.el b/eglot.el
index 22a8aef..fd37bb7 100644
--- a/eglot.el
+++ b/eglot.el
@@ -2288,41 +2288,41 @@ Buffer is displayed with `display-buffer', which obeys
 `display-buffer-alist' & friends."
   :type 'boolean)
 
+(defun eglot--first-line-of-doc (string)
+  (truncate-string-to-width
+   (replace-regexp-in-string "\\(.*\\)\n.*" "\\1" string)
+   (frame-width) nil nil "..."))
+
 (defun eglot--update-doc (string hint)
   "Put updated documentation STRING where it belongs.
 Honours `eglot-put-doc-in-help-buffer'.  HINT is used to
 potentially rename EGLOT's help buffer.  If STRING is nil, the
 echo area cleared of any previous documentation."
-  (cond ((and string
-              (or (eq t eglot-put-doc-in-help-buffer)
-                  (and eglot-put-doc-in-help-buffer
-                       (funcall eglot-put-doc-in-help-buffer string))))
-         (with-current-buffer (eglot--help-buffer)
-           (let ((inhibit-read-only t)
-                 (name (format "*eglot-help for %s*" hint)))
-             (unless (string= name (buffer-name))
-               (rename-buffer (format "*eglot-help for %s*" hint))
-               (erase-buffer)
-               (insert string)
-               (goto-char (point-min)))
-             (if eglot-auto-display-help-buffer
-                 (display-buffer (current-buffer))
-               (unless (get-buffer-window (current-buffer))
-                 (eglot--message
-                  "%s\n(...truncated. Full help is in `%s')"
-                  (truncate-string-to-width
-                   (replace-regexp-in-string "\\(.*\\)\n.*" "\\1" string)
-                   (frame-width) nil nil "...")
-                  (buffer-name eglot--help-buffer))))
-             (help-mode))))
-        (eldoc-echo-area-use-multiline-p
-         (eldoc-message string))
-        (t
-         (eldoc-message
-          (and string
-               (if (string-match "\n" string)
-                   (substring string (match-end 0))
-                 string))))))
+  (when string
+    (cond ((or (eq t eglot-put-doc-in-help-buffer)
+               (and eglot-put-doc-in-help-buffer
+                    (funcall eglot-put-doc-in-help-buffer string)))
+           (with-current-buffer (eglot--help-buffer)
+             (let ((inhibit-read-only t)
+                   (name (format "*eglot-help for %s*" hint)))
+               (unless (string= name (buffer-name))
+                 (rename-buffer (format "*eglot-help for %s*" hint))
+                 (erase-buffer)
+                 (insert string)
+                 (goto-char (point-min)))
+               (if eglot-auto-display-help-buffer
+                   (display-buffer (current-buffer))
+                 (unless (get-buffer-window (current-buffer))
+                   (eglot--message
+                    "%s\n(...truncated. Full help is in `%s')"
+                    (eglot--first-line-of-doc string)
+                    (buffer-name eglot--help-buffer))))
+               (help-mode))))
+          (eldoc-echo-area-use-multiline-p
+           (eldoc-message string))
+          (t
+           (eldoc-message
+            (eglot--first-line-of-doc string))))))
 
 (defun eglot-eldoc-function ()
   "EGLOT's `eldoc-documentation-function' function."

```

muffinmad:
1. Full hover info must be shown only if `eldoc-echo-area-use-multiline-p` is `t`. Otherwise show first line.
2. Even if we show full hover info (`eldoc-echo-area-use-multiline-p` is `t`), STRING must be truncated to fit the echo area. This way first lines of the huge hover info will be shown in the echo area instead of last lines.
3. Why use `eglot-message` to show hover info when `eglot-put-doc-in-help-buffer` is non-nil? Why `eldoc-message` can't be used?

PR #459, opened 2020-05-06
Possibility to show full first line of hover info

muffinmad:
Let's continue our journey to perfect eldoc message! :)

This PR provides the possibility to show a full first line of the hover/signature info.

**Current behavior**

|`eldoc-echo-area-use-multiline-p` value|What part of the hover info we see in the echo area|
|-|-|
|`nil`|First line truncated to the frame width<br>*"Hey, where is the rest of the signature info? I just finished with the tenth arg of the func and want to know what the eleventh one is about!"*|
|non-nil|Last lines<br>*"Why, Eglot, why?"*|

**Proposed behavior**

|`eldoc-echo-area-use-multiline-p` value|What part of the hover info we see in the echo area|
|-|-|
|`nil`|First line truncated to the frame width<br>*"The echo area occupy only one line and doesn't break my shiny windows layout. Whoo-hoo!"*|
|`t`|Top lines<br>*"I can see as much hover info as the echo area can fit. Very informative! Awesome!"*|
|non-nil|Full first line<br>*"OMG this is the most convenient way to display hover and (especially) signature info! Great job, Eglot!"*|

**Bonus**

Honor `eglot--message` prefix length when truncating message.

`(setq eglot-put-doc-in-help-buffer t)`

The echo area content:

*Current behavior*

```
[eglot] datetime(year: int, month: int, day: int, hour: int=..., minute: int=...
, sec...
(...truncated. Full help is in `*eglot-help for datetime*')
```

*Proposed behavior*

```
[eglot] datetime(year: int, month: int, day: int, hour: int=..., minute: int=...
(...truncated. Full help is in `*eglot-help for datetime*')
```

Thanks!

### DIFF
--- a/eglot.el
+++ b/eglot.el
@@ -2281,24 +2281,47 @@ is not active."
               (setq-local nobreak-char-display nil)))
         (display-local-help)))))
 
-(defun eglot-doc-too-large-for-echo-area (string)
-  "Return non-nil if STRING won't fit in echo area.
-Respects `max-mini-window-height' (which see)."
-  (let ((max-height
-         (cond ((floatp max-mini-window-height) (* (frame-height)
-                                                   max-mini-window-height))
-               ((integerp max-mini-window-height) max-mini-window-height)
-               (t 1))))
-    (> (cl-count ?\n string) max-height)))
+(cl-defun eglot-doc-too-large-for-echo-area
+    (string &optional (height max-mini-window-height))
+  "Return non-nil if STRING won't fit in echo area of height HEIGHT.
+HEIGHT defaults to `max-mini-window-height' (which see) and is
+interpreted like that variable.  If non-nil, the return value is
+the number of lines available."
+  (let ((available-lines (cl-typecase height
+                           (float (truncate (* (frame-height) height)))
+                           (integer height)
+                           (t 1))))
+    (when (> (1+ (cl-count ?\n string)) available-lines)
+      available-lines)))
+
+(cl-defun eglot--truncate-string (string height &optional (width (frame-width)))
+  "Return as much from STRING as fits in HEIGHT and WIDTH.
+WIDTH, if non-nil, truncates last line to those columns."
+  (cl-flet ((maybe-trunc
+             (str) (if width (truncate-string-to-width str width
+                                                       nil nil "...")
+                     str)))
+    (cl-loop
+     repeat height
+     for i from 1
+     for break-pos = (cl-position ?\n string)
+     for (line . rest) = (and break-pos
+                              (cons (substring string 0 break-pos)
+                                    (substring string (1+ break-pos))))
+     concat (cond (line (if (= i height) (maybe-trunc line) (concat line "\n")))
+                  (t (maybe-trunc string)))
+     while rest do (setq string rest))))
 
 (defcustom eglot-put-doc-in-help-buffer
+  ;; JT@2020-05-21: TODO: this variable should be renamed and the
+  ;; decision somehow be in eldoc.el itself.
   #'eglot-doc-too-large-for-echo-area
   "If non-nil, put \"hover\" documentation in separate `*eglot-help*' buffer.
 If nil, use whatever `eldoc-message-function' decides, honouring
 `eldoc-echo-area-use-multiline-p'.  If t, use `*eglot-help*'
-unconditionally.  If a function, it is called with the docstring
-to display and should a boolean producing one of the two previous
-values."
+unconditionally.  If a function, it is called with the
+documentation string to display and returns a generalized boolean
+interpreted as one of the two preceding values."
   :type '(choice (const :tag "Never use `*eglot-help*'" nil)
                  (const :tag "Always use `*eglot-help*'" t)
                  (function :tag "Ask a function")))
@@ -2308,11 +2331,6 @@ values."
 Buffer is displayed with `display-buffer', which obeys
 `display-buffer-alist' & friends."
   :type 'boolean)
-
-(defun eglot--first-line-of-doc (string)
-  (truncate-string-to-width
-   (replace-regexp-in-string "\\(.*\\)\n.*" "\\1" string)
-   (frame-width) nil nil "..."))
 
 (defun eglot--update-doc (string hint)
   "Put updated documentation STRING where it belongs.
@@ -2336,16 +2354,22 @@ documentation.  Honour `eglot-put-doc-in-help-buffer',
              (if eglot-auto-display-help-buffer
                  (display-buffer (current-buffer))
                (unless (get-buffer-window (current-buffer))
+                 ;; This prints two lines.  Should it print 1?  Or
+                 ;; honour max-mini-window-height?
                  (eglot--message
                   "%s\n(...truncated. Full help is in `%s')"
-                  (eglot--first-line-of-doc string)
+                  (eglot--truncate-string string 1 (- (frame-width) 8))
                   (buffer-name eglot--help-buffer))))
              (help-mode))))
-        (eldoc-echo-area-use-multiline-p
-         ;; Can't really honour non-t non-nil values if this var
-         (eldoc-message string))
+        ((eq eldoc-echo-area-use-multiline-p t)
+         (if-let ((available (eglot-doc-too-large-for-echo-area string)))
+             (eldoc-message (eglot--truncate-string string available))
+           (eldoc-message string)))
+        ((eq eldoc-echo-area-use-multiline-p 'truncate-sym-name-if-fit)
+         (eldoc-message (eglot--truncate-string string 1 nil)))
         (t
-         (eldoc-message (eglot--first-line-of-doc string)))))
+         ;; Can't (yet?) honour non-t non-nil values of this var
+         (eldoc-message (eglot--truncate-string string 1)))))
 
 (defun eglot-eldoc-function ()
   "EGLOT's `eldoc-documentation-function' function."
--- a/eglot-tests.el
+++ b/eglot-tests.el
@@ -573,6 +573,32 @@ def foobazquuz(d, e, f): pass
        (while (not eldoc-last-message) (accept-process-output nil 0.1))
        (should (string-match "^exit" eldoc-last-message))))))
 
+(ert-deftest hover-multiline-doc-locus ()
+  "Test if suitable amount of lines of hover info are shown."
+  (skip-unless (executable-find "pyls"))
+  (eglot--with-fixture
+      `(("project" . (("hover-first.py" . "from datetime import datetime")))
+        (eglot-put-doc-in-help-buffer nil)
+        ,@eglot--tests--python-mode-bindings)
+    (with-current-buffer
+        (eglot--find-file-noselect "project/hover-first.py")
+      (should (eglot--tests-connect))
+      (goto-char (point-max))
+      ;; one-line
+      (setq eldoc-last-message nil)
+      (setq-local eldoc-echo-area-use-multiline-p nil)
+      (eglot-eldoc-function)
+      (while (not eldoc-last-message) (accept-process-output nil 0.1))
+      (should (string-match "datetime" eldoc-last-message))
+      (should (not (cl-find ?\n eldoc-last-message)))
+      ;; multi-line
+      (setq eldoc-last-message nil)
+      (setq-local eldoc-echo-area-use-multiline-p t)
+      (eglot-eldoc-function)
+      (while (not eldoc-last-message) (accept-process-output nil 0.1))
+      (should (string-match "datetime" eldoc-last-message))
+      (should (cl-find ?\n eldoc-last-message)))))
+
 (ert-deftest python-autopep-formatting ()
   "Test formatting in the pyls python LSP.
 pyls prefers autopep over yafp, despite its README stating the contrary."


muffinmad:
This can probably fix #117

joaotavora:
:-) nice job 👏 👏 👏 

I think the idea is good, but it doens't follow the semantics of the `eldoc` variable exactly.  However, that wasn't happening anyway, and I think I can live with that, especially given such a convincing presentation.  We can look at the implementation tomorrow.

joaotavora:
Hi @muffinmad , I did some adjustments to this PR, to join some duplicated logic, improving efficiency, but hopefully keeping the behaviour you want.  I'm in a hurry, so I couldn't even test, but it shouldn't be very far off.  Have a look when you can.

muffinmad:
Alongside with failed newly added test following error occurred.

```python
from datetime import datetime
```

With point at the end of that line:

```
Debugger entered--Lisp error: (wrong-type-argument number-or-marker-p nil)
  1-(nil)
  (setq --cl-var-- (1- --cl-var--))
  (>= (setq --cl-var-- (1- --cl-var--)) 0)
  (and (>= (setq --cl-var-- (1- --cl-var--)) 0) (progn (setq break-pos (cl-position 10 string)) (setq rest (and break-pos (cons (substring string 0 break-pos) (substring string (1+ break-pos)))) line (car-safe (prog1 rest (setq rest (cdr rest))))) (if line (progn (setq --cl-var-- (concat --cl-var-- (if (= i 1) "" "\n"))) (setq --cl-var-- (concat --cl-var-- (if (= i height) (truncate-string-to-width line width nil nil "...") line)))) (setq --cl-var-- (concat --cl-var-- string))) rest))
  (while (and (>= (setq --cl-var-- (1- --cl-var--)) 0) (progn (setq break-pos (cl-position 10 string)) (setq rest (and break-pos (cons (substring string 0 break-pos) (substring string (1+ break-pos)))) line (car-safe (prog1 rest (setq rest (cdr rest))))) (if line (progn (setq --cl-var-- (concat --cl-var-- (if ... "" "\n"))) (setq --cl-var-- (concat --cl-var-- (if ... ... line)))) (setq --cl-var-- (concat --cl-var-- string))) rest)) (setq string rest) (setq i (+ i 1)) (setq --cl-var-- nil))
  (let* ((--cl-var-- height) (i 1) (break-pos nil) (rest nil) (line nil) (--cl-var-- "") (--cl-var-- t)) (while (and (>= (setq --cl-var-- (1- --cl-var--)) 0) (progn (setq break-pos (cl-position 10 string)) (setq rest (and break-pos (cons (substring string 0 break-pos) (substring string ...))) line (car-safe (prog1 rest (setq rest ...)))) (if line (progn (setq --cl-var-- (concat --cl-var-- ...)) (setq --cl-var-- (concat --cl-var-- ...))) (setq --cl-var-- (concat --cl-var-- string))) rest)) (setq string rest) (setq i (+ i 1)) (setq --cl-var-- nil)) --cl-var--)
  (progn (if --cl-rest-- (signal 'wrong-number-of-arguments (list 'eglot--truncate-string (+ 3 (length --cl-rest--))))) (let* ((--cl-var-- height) (i 1) (break-pos nil) (rest nil) (line nil) (--cl-var-- "") (--cl-var-- t)) (while (and (>= (setq --cl-var-- (1- --cl-var--)) 0) (progn (setq break-pos (cl-position 10 string)) (setq rest (and break-pos (cons ... ...)) line (car-safe (prog1 rest ...))) (if line (progn (setq --cl-var-- ...) (setq --cl-var-- ...)) (setq --cl-var-- (concat --cl-var-- string))) rest)) (setq string rest) (setq i (+ i 1)) (setq --cl-var-- nil)) --cl-var--))
  (let* ((width (if --cl-rest-- (car-safe (prog1 --cl-rest-- (setq --cl-rest-- (cdr --cl-rest--)))) (frame-width)))) (progn (if --cl-rest-- (signal 'wrong-number-of-arguments (list 'eglot--truncate-string (+ 3 (length --cl-rest--))))) (let* ((--cl-var-- height) (i 1) (break-pos nil) (rest nil) (line nil) (--cl-var-- "") (--cl-var-- t)) (while (and (>= (setq --cl-var-- (1- --cl-var--)) 0) (progn (setq break-pos (cl-position 10 string)) (setq rest (and break-pos ...) line (car-safe ...)) (if line (progn ... ...) (setq --cl-var-- ...)) rest)) (setq string rest) (setq i (+ i 1)) (setq --cl-var-- nil)) --cl-var--)))
  eglot--truncate-string(#("datetime(year: int, month: int, day: int, hour: in..." 15 18 (face font-lock-builtin-face) 27 30 (face font-lock-builtin-face) 37 40 (face font-lock-builtin-face) 48 51 (face font-lock-builtin-face) 65 68 (face font-lock-builtin-face) 82 85 (face font-lock-builtin-face) 104 107 (face font-lock-builtin-face)) nil)
  (eldoc-message (eglot--truncate-string string (eglot-doc-too-large-for-echo-area string)))
  (cond ((null string) (eldoc-message nil)) ((or (eq t eglot-put-doc-in-help-buffer) (and eglot-put-doc-in-help-buffer (funcall eglot-put-doc-in-help-buffer string))) (save-current-buffer (set-buffer (eglot--help-buffer)) (let ((inhibit-read-only t) (name (format "*eglot-help for %s*" hint))) (if (string= name (buffer-name)) nil (rename-buffer (format "*eglot-help for %s*" hint)) (erase-buffer) (insert string) (goto-char (point-min))) (if eglot-auto-display-help-buffer (display-buffer (current-buffer)) (if (get-buffer-window (current-buffer)) nil (eglot--message "%s\n(...truncated. Full help is in `%s')" (eglot--truncate-string string 1 ...) (buffer-name eglot--help-buffer)))) (help-mode)))) ((eq eldoc-echo-area-use-multiline-p t) (eldoc-message (eglot--truncate-string string (eglot-doc-too-large-for-echo-area string)))) (t (eldoc-message (eglot--truncate-string string 1))))
  eglot--update-doc(#("datetime(year: int, month: int, day: int, hour: in..." 15 18 (face font-lock-builtin-face) 27 30 (face font-lock-builtin-face) 37 40 (face font-lock-builtin-face) 48 51 (face font-lock-builtin-face) 65 68 (face font-lock-builtin-face) 82 85 (face font-lock-builtin-face) 104 107 (face font-lock-builtin-face)) #("datetime" 0 8 (fontified t)))
  (save-current-buffer (set-buffer buffer) (eglot--update-doc (and (not (seq-empty-p contents)) (eglot--hover-info contents range)) thing-at-point))
  (progn (save-current-buffer (set-buffer buffer) (eglot--update-doc (and (not (seq-empty-p contents)) (eglot--hover-info contents range)) thing-at-point)))
  (if (or (get-buffer-window buffer) (ert-running-test)) (progn (save-current-buffer (set-buffer buffer) (eglot--update-doc (and (not (seq-empty-p contents)) (eglot--hover-info contents range)) thing-at-point))))
  (if sig-showing nil (if (or (get-buffer-window buffer) (ert-running-test)) (progn (save-current-buffer (set-buffer buffer) (eglot--update-doc (and (not (seq-empty-p contents)) (eglot--hover-info contents range)) thing-at-point)))))
  (progn (eglot--check-object 'Hover object-once (memq 'enforce-required-keys eglot-strict-mode) (memq 'disallow-non-standard-keys eglot-strict-mode) (memq 'check-types eglot-strict-mode)) (if sig-showing nil (if (or (get-buffer-window buffer) (ert-running-test)) (progn (save-current-buffer (set-buffer buffer) (eglot--update-doc (and (not ...) (eglot--hover-info contents range)) thing-at-point))))))
  (let* ((--cl-rest-- object-once) (contents (car (cdr (plist-member --cl-rest-- ':contents)))) (range (car (cdr (plist-member --cl-rest-- ':range))))) (progn (eglot--check-object 'Hover object-once (memq 'enforce-required-keys eglot-strict-mode) (memq 'disallow-non-standard-keys eglot-strict-mode) (memq 'check-types eglot-strict-mode)) (if sig-showing nil (if (or (get-buffer-window buffer) (ert-running-test)) (progn (save-current-buffer (set-buffer buffer) (eglot--update-doc (and ... ...) thing-at-point)))))))
  (let ((object-once jsonrpc-lambda-elem12)) (let* ((--cl-rest-- object-once) (contents (car (cdr (plist-member --cl-rest-- ':contents)))) (range (car (cdr (plist-member --cl-rest-- ':range))))) (progn (eglot--check-object 'Hover object-once (memq 'enforce-required-keys eglot-strict-mode) (memq 'disallow-non-standard-keys eglot-strict-mode) (memq 'check-types eglot-strict-mode)) (if sig-showing nil (if (or (get-buffer-window buffer) (ert-running-test)) (progn (save-current-buffer (set-buffer buffer) (eglot--update-doc ... thing-at-point))))))))
  (closure ((thing-at-point . #("datetime" 0 8 (fontified t))) (sig-showing) (position-params :textDocument (:uri "file:///Users/mad/workspace/test/test.py") :position (:line 0 :character 29)) (server . #<eglot-lsp-server eglot-lsp-server-1fe76b40c8b4>) (buffer . #<buffer test.py>) revert-buffer-preserve-modes eglot--managed-mode company-tooltip-align-annotations company-backends markdown-fontify-code-blocks-natively t) (jsonrpc-lambda-elem12) (let ((object-once jsonrpc-lambda-elem12)) (let* ((--cl-rest-- object-once) (contents (car (cdr ...))) (range (car (cdr ...)))) (progn (eglot--check-object 'Hover object-once (memq 'enforce-required-keys eglot-strict-mode) (memq 'disallow-non-standard-keys eglot-strict-mode) (memq 'check-types eglot-strict-mode)) (if sig-showing nil (if (or ... ...) (progn ...)))))))((:contents [(:value "datetime(year: int, month: int, day: int, hour: in..." :language "python") "datetime(year, month, day[, hour[, minute[, second..."]))
  jsonrpc-connection-receive(#<eglot-lsp-server eglot-lsp-server-1fe76b40c8b4> (:jsonrpc "2.0" :id 6 :result (:contents [(:value "datetime(year: int, month: int, day: int, hour: in..." :language "python") "datetime(year, month, day[, hour[, minute[, second..."])))
  jsonrpc--process-filter(#<process EGLOT (test/python-mode)> "Content-Length: 460\15\nContent-Type: application/vsc...")
```

joaotavora:
Nice, thanks for the test and the simple recipe.  I'll look at this asap

fbergroth:
There seems to be a bug in `eglot--truncate-string` if there is no trailing newline:
```(eglot--truncate-string "aa\nbb" 2) ;; => "aabb"```

**Comment by joaotavora** [comment]
Fredrik Bergroth <notifications@github.com> writes:

> There seems to be a bug in eglot--truncate-string if there is no trailing newline:
> (eglot--truncate-string "aa\nbb" 2) ;; => "aabb"

thanks!

**Comment by joaotavora** [comment]
Hi @muffinmad and @fbergroth.  I fixed both problems, and enhanced the test a little bit.

@muffinmad if this works like you wanted I was thinking of squashing this to a commit and pushing with you as the Co-author. Or the other way around, you the author me co-author. Whichever you prefer.

File: eglot.el [hunk 1]
```diff
@@ -2336,16 +2355,20 @@ documentation.  Honour `eglot-put-doc-in-help-buffer',
...
-        (eldoc-echo-area-use-multiline-p
-         ;; Can't really honour non-t non-nil values if this var
-         (eldoc-message string))
+        ((eq eldoc-echo-area-use-multiline-p t)
+         (if-let ((available (eglot-doc-too-large-for-echo-area string)))
+             (eldoc-message (eglot--truncate-string string available))
+           (eldoc-message string)))
         (t
-         (eldoc-message (eglot--first-line-of-doc string)))))
+         ;; Can't (yet?) honour non-t non-nil values if this var
```
dandavison, line 2370: Should this read "of this var"?

eglot.el
```diff
@@ -2281,24 +2281,48 @@ is not active."
...
-         (cond ((floatp max-mini-window-height) (* (frame-height)
-                                                   max-mini-window-height))
-               ((integerp max-mini-window-height) max-mini-window-height)
-               (t 1))))
-    (> (cl-count ?\n string) max-height)))
+(cl-defun eglot-doc-too-large-for-echo-area
+    (string &optional (height max-mini-window-height))
+  "Return non-nil if STRING won't fit in echo area of height HEIGHT.
+HEIGHT defaults to `max-mini-window-height' (which see) and is
+interpret like that variable.  If non-nil, the return value is
```
dandavison, line 2288: typo: "interpreted"

eglot.el
joaotavora: fixed

eglot.el
joaotavora: fixed

muffinmad:
> if this works like you wanted

Please take a look at the differences between the original PR.

All tests are done with

```elisp
(setq eglot-put-doc-in-help-buffer nil)
(modify-frame-parameters nil '((width . 70) (height . 20)))
```

**1.** `(setq eldoc-echo-area-use-multiline-p nil)`

No problem here, the first line of the hover info truncated to fame width is shown.

**2.** `(setq eldoc-echo-area-use-multiline-p 'truncate-sym-name-if-fit)`

First line of the hover info truncated to frame width is show. Can we show full first line here?

Screenshot 2020-05-24 at 23 13 37

*New PR version is on the left*

**3.** `(setq eldoc-echo-area-use-multiline-p t)`

First lines of the hover info are hidden.

Screenshot 2020-05-24 at 23 15 36

*New PR version is on the left*

> I was thinking of squashing this to a commit and pushing with you as the Co-author. Or the other way around, you the author me co-author. Whichever you prefer.

I'm totally fine with the first option.  It's turned to be your PR now ;-)

joaotavora:
> I'm totally fine with the first option. It's turned to be your PR now ;-)

Yes, I'm sorry. Though I didnt' erase your commit (dont' lose it yoruslef).

Anyway I'm not going to push until this does exactly what you want. What matters is the behaviour, and then we make the implementation as tight as possible.

joaotavora:
OK, this should be fixed (both problems you pointed out in 2 and 3). Have a last look and I'll push tomorrow to close this.  And thanks very much!

muffinmad:
> OK, this should be fixed (both problems you pointed out in 2 and 3).

Case number 2 works as expected now, but for 3 the first lines of the hover info are still hidden:

```elisp
(modify-frame-parameters nil '((width . 70)))
(setq eldoc-echo-area-use-multiline-p t)
```

Screenshot 2020-05-25 at 09 47 11

joaotavora:
I think this third problem should be something solved in `eldoc` perhaps by setting `message-truncate-lines` to a non nil value. I.e. eglot (or other eldoc users) shouldn't be making complex math to handle this case.  Perhaps @dgutov or @monnier would like to weigh in.  By the way @muffinmad, you can make "pull requests" on eldoc.el by forking https://github.com/emacs-mirror/emacs and doing your changes there.  It's easy to convert that to a patch that we can push to core, perhaps after discussing it in the Emacs bug tracker.  I suggest you get comfortable with this idea ;-)

joaotavora:
In the meantime, I've pushed this version to master, since it's much nicer than what we had before.

dgutov:
So far we've been leaving that responsibility to `eglot-documentation-function` (binding `message-truncate-lines` to t). That doesn't need complex math.

Another thing to try is binding `resize-mini-windows` to nil, I think.

joaotavora:
Indeed, I tried that, but I think something in eldoc itself will then call
eldoc-message again with the last recorded message, but without the binding.

Anyway this is just a middle point before actually fixing the need for
eldoc clients to call eldoc-message directly.

dgutov:
> Indeed, I tried that, but I think something in eldoc itself will then call
eldoc-message again with the last recorded message, but without the binding.

Ah, hm, my bad. Indeed, you can't set the binding this way.

But looking at ElDoc's sources, have you tried the `eldoc-echo-area-use-multiline-p` defcustom?

dgutov:
And by "tried", I mean "looked at its uses".

It's how enforced automatically (leaving it to the `e-d-f` functions), but that part could be improved.

joaotavora:
> It's not enforced automatically (leaving it to the e-d-f functions), but that part could be improved.

Indeed, I think so too.  In fact, the only thing that theoretically can't be enforced by `eldoc` directly is the `truncate-sym-name-if-fit` value, because it doesn't have a clue what the sym name is.  All else could be handled by it, I think.

dgutov:
Speaking of this, though:

> fixing the need for eldoc clients to call eldoc-message directly

Is there a particular plan? How would `eldoc-message` be different from plain `message` in this usage?

joaotavora:
> Is there a particular plan?

The plan is for the user of eldoc not to call `eldoc-message` or `message` at all, it's being able to returning some other non-nil value from `eldoc-documentation-function` whereby eglot says to eldoc:

> "sorry, eldoc, I don't have the string just yet but I promise to call back the function you put in this special variable `xxx` when I do (because unfortunately your can't pass me any args without breaking backward compatibility)".

If you're not a fan of the "special variable that holds function" thing, we can create a `eldoc-deferred-documentation-function` that does take one arg, said callback function.  We call that before we call `eldoc-documentation-function` and if it returns non-nil. If it returns a string, we assume it's a docstring to display immediately.  This is very similar to how Flymake works already.

monnier:
>> "sorry, eldoc, I don't have the string just yet but I promise to call back
>> the function you put in this special variable `xxx` when I do (because
>> unfortunately your can't pass me any args without breaking backward
>> compatibility)".
>
> If you're not a fan of the "special variable that holds function" thing, we
> can create a `eldoc-deferred-documentation-function` that does take one arg,
> said callback function.  We call that before we call
> `eldoc-documentation-function` and if it returns non-nil. If it returns
> a string, we assume it's a docstring to display immediately.  This is very
> similar to how Flymake works already.

I understand that in theory, eldoc may want each call to go
through a different callback, but in practice, I can't think of a good
reason for that.

So I think a much simpler API is to have `eldoc-documentation-function*s*`
return a special value meaning "will call you later" and that later call
is to a global function `eldoc-documentation-callback`.

In the current state of things, maybe `eldoc-documentation-callback`
wouldn't do much more than just call `eldoc-message`.
But I guess would could make it accept an additional argument which
indicates who calls it (either the name of the function that's placed
on `eldoc-documentation-functions` or some arbitrary value which might
be the same value returned to mean "will call you later").


        Stefan

joaotavora:
> So I think a much simpler API is to have `eldoc-documentation-function*s*`

Actually, I disagree this is simpler.  We'd be doing the "var x is obsolete" dance again.  This might be simpler than you're making it:

What you are suggesting (`eldoc-documentation-callback`) is my first suggestion exactly, the special variable one.  No problem with that (and so I'm gonna start working on that one.)  I do think it's quite orthogonal to performing the other change (`eldoc-documentation-function` to `eldoc-documentation-functions`).  

Of course, if later on we do want the "different callback for each client", then we can make that rename.  But I _also_ agree with you that that's improbable (though not impossible).

joaotavora:
Oh I see there's already `eldoc-documentation-functions`. My bad.  So we've already done that dance.  No problem I guess, shouldn't invalidate my patch.

monnier:
>> So I think a much simpler API is to have `eldoc-documentation-function*s*`
> Actually, I disagree this is simpler.  We'd be doing the "var x is obsolete"
> dance again.  This might be simpler than you're making it:

I didn't mean that adding an `s` makes it simpler.

I only highlighted the `s` because I wanted to be precise about what I'm
talking about (since we have both `eldoc-documentation-function` and
`eldoc-documentation-functions`).  But the "simpler" argument was about
calling a constant&global function rather than going through a "special
variable" or having to call some special function before returning the
"will call you later" value.

> What you are suggesting (`eldoc-documentation-callback`) is my first
> suggestion exactly, the special variable one.

Not sure why you think so: I mean it as a function name, not a variable name.

> No problem with that (and so I'm gonna start working on that one.)

Thanks ;-)

> I do think it's quite orthogonal to performing the other change
> (`eldoc-documentation-function` to `eldoc-documentation-functions`).  

Hmm... I'm beginning to get the impression that you missed
commit c0fcbd2c119b8418855f0931aceefbef717c5e53 `master`:

    ** ElDoc
    
    *** New hook 'eldoc-documentation-functions' to be used for registering
    doc string functions.  This makes the results of all doc string
    functions accessible to the user through the existing single function hook
    'eldoc-documentation-function'.
    
    *** 'eldoc-documentation-function' is now a user option.
    Modes should use the new hook instead of this user option to register
    their backends.


-- Stefan

joaotavora:
> Not sure why you think so: I mean it as a function name, not a variable name.

`eldoc-documentation-callback` is a function, not a variable?  If so, quite a strange name for a function.  Then what's the difference between that and calling `eldoc-message` directly?  I guess just the return value of the things eglot is going to put in `eldoc-documentation-functions`.  I'd rather use the variable though, it gives us more leeway for handling quirky cases later on, like properly supporting the `eldoc-documentation-compose` case.

By the way, this is unrelated: _some_ handling of `eldoc-echo-blablabla-multiline` is already done by eldoc.el directly I see.  Docstrings should be upgraded.




> Hmm... I'm beginning to get the impression that you missed
> commit c0fcbd2c119b8418855f0931aceefbef717c5e53 `master`:

Yes, as I said before i missed it.  It makes things more complicated.

dgutov:
> So I think a much simpler API is to have `eldoc-documentation-function*s*`
return a special value meaning "will call you later" and that later call
is to a global function `eldoc-documentation-callback`.

Why doesn't it return some "future" or "promise" value instead? So that Eldoc sets up all the necessary callbacks itself.

That should make the API much cleaner.

We'd only need to add said promise/future type to the standard library. :-)

joaotavora:
> We'd only need to add said promise/future type to the standard library. :-)

If you can show a patch that does all this and demonstrates the cleanliness of this approach and keeps backward compatibility, I'm all for it.  I'm going old school callback route (which is basically future with a less fancy name).

dgutov:
The exact patch would depend on the presence of said Future type. And on whatever extra requirements we might want for it. Because having a standard Future, documented in one place, and used from several places, will be beneficial for us.

If we're not ready for it, then here's a trivial example of usage of an ad-hoc future that wouldn't require for `eldoc-documentation-functions` element to know about `eldoc-documentation-callback`: https://github.com/company-mode/company-mode/blob/36e650acd681a3a15a96e3c074471e86bdeeb4d8/company-clang.el#L376-L377

If the idea is not obvious from that snippet, I can cook up a full patch, no problem.

joaotavora:
That's basically what I'm doing now.  Wait for my patch please.  Or cook up your own and if they both solve the same problem, great, we'll take the pick of the litter.

joaotavora:
Gotta love the "they are Legion" comment in eldoc.el :rofl:

monnier:
> `eldoc-documentation-callback` is a function, not a variable?  If so, quite
> a strange name for a function.

Feel free to use something better.

> Then what's the difference between that and calling `eldoc-message` directly?

[ I have a feeling of déjà vu!  ]

`eldoc-message` (despite the absence of a "--") is an
internal function.  We could "redirect it", but I think at this point we
might as well define a new function to be used for the call back.
If in the end it can be merged with `eldoc-message`, then that's fine as well.

In any case, there may be differences, especially now with the *s*
which means that we need to combine the output from various functions
before we can pass it to `eldoc-message`.

> I guess just the return value of the things eglot is going to put in
> `eldoc-documentation-functions`.  I'd rather use the variable though,
> it gives us more leeway for handling quirky cases later on, like
> properly supporting the `eldoc-documentation-compose` case.

But that should all be hidden behind `eldoc-documentation-callback`
(the functions placed on the *s* shouldn't have to worry about that).

> By the way, this is unrelated: _some_ handling of
> `eldoc-echo-blablabla-multiline` is already done by eldoc.el directly
> I see.  Docstrings should be upgraded.

Quite likely.  I think there's a chance that the code should be changed
as well (and not necessarily to reflect the doc, but to provide a more
sane/regular/predictable behavior).

>> Hmm... I'm beginning to get the impression that you missed
>> commit c0fcbd2c119b8418855f0931aceefbef717c5e53 `master`:
> Yes, as I said before i missed it.  It makes things more complicated.

FWIW, while I think the idea behind the change is right, I'm beginning
to suspect that a different approach might be needed (since we will need
to combine synchronous outputs with asynchronously outputs, so I suspect
we won't be able to hide all that within
`eldoc-documentation-function`).

FWIW, the current code also has a bit of a backward compatibility
problem with packages that set `eldoc-documentation-function` instead of
`eldoc-documentation-functions`.  So this might be an opportunity to fix
this backward compatibility as well before it's "anointed" in the
Emacs-28 release.


        Stefan

monnier:
>> So I think a much simpler API is to have `eldoc-documentation-function*s*`
> return a special value meaning "will call you later" and that later call
> is to a global function `eldoc-documentation-callback`.
>
> Why doesn't it return some "future" or "promise" value instead? So that
> Eldoc sets up all the necessary callbacks itself.

Fine by me


        Stefan

joaotavora:
> But that should all be hidden behind `eldoc-documentation-callback`
> (the functions placed on the *s* shouldn't have to worry about that).

Oh sure, but if we can't give each member of those functions a different 
callback function, then it's harder to combine the results sanely.  In hindsight
it was bad to make elements `eldoc-documentation-functions` argless functions. 
You could have followed a similar path to `flymake-diagnostic-functions` for 
what is a completely analogous problem.  (yes Dmitry, Flymake could _also_ 
use futures :-) ).  Maybe you still can, this API hasn't really been out for longer
than a few weeks in GNU ELPA, I doubt anyone is using it.

Anyway, the current patch I'm preparing doens't suppport the async case and
`eldoc-documentation-compose` yet.  It _could_ support it the future, and work 
much the same way that flymake successfully mixes sync and async sources.
But that's a "nice to have", IMO, I don't see any demand for it.

monnier:
>> But that should all be hidden behind `eldoc-documentation-callback`
>> (the functions placed on the *s* shouldn't have to worry about that).
> Oh sure, but if we can't give each member of those functions a different
> callback function, then it's harder to combine the results sanely.  In hindsight
> it was bad to make elements `eldoc-documentation-functions` argless functions.

I'm not sure what information they'd need to do their job.

OK, let's say we use futures.

So we say that `eldoc-documentation-functions` can return futures.
The futures API I'd imagine it here could look like:

    (cl-defstruct (eldoc-future
                   (conc-name eldoc-future--)
                   (:constructor eldoc-future-make ()))
     (value eldoc-future--unset)
     callback)

    (defun eldoc-future-set (f v)
      (cl-assert (eq (eldoc-future--value f) 'eldoc-future--unset))
      (setf (eldoc-future--value f) v)
      (when (eldoc-future--callback f)
        (funcall (eldoc-future--callback f) v)))

    (defun eldoc-future-set-callback (f c)
      (cl-assert (null (eldoc-future--callback f)))
      (setf (eldoc-future--callback f) c)
      (unless (eq (eldoc-future--value f) 'eldoc-future--unset)
        (funcall c (eldoc-future--value f))))

So the backend starts by calling `eldoc-future-make` and returning
that object.  Then eldoc calls `eldoc-future-set-callback` on it, which
sets the callback.  Then in a process-filter/sentinel the backend calls
`eldoc-future-set` on it, thus running the callback.

The same interface would work for synchronous backends, where they call
`eldoc-future-set` before returning the object.

Then we write a `eldoc-future-do` macro that turns

    (eldoc-future-let ((v FOO))
      BAR)

into

    (eldoc-future-set-callback FOO (lambda (v) BAR))

> You could have followed a similar path to `flymake-diagnostic-functions` for 
> what is a completely analogous problem.

You mean we'd pass the callback function as an argument?
That's a valid option as well, indeed.

I think what I like about futures is that it opens up the possibility to
extend the API a bit more easily.  E.g. adding the possibility to abort
a future.

> Maybe you still can, this API hasn't really been out for longer than
> a few weeks in GNU ELPA, I doubt anyone is using it.

I definitely consider this API as "not yet released".


        Stefan

joaotavora:
I don't have anything against this futuristic ambitious future stuff, but I definitely don't have time to discuss it, and this problem needs some fixing in the short term (both here, in SLY, and probably also SLIME). So I'm going to propose something much closer to what `flymake-diagnostic-functions` already does, which is the patch I attach here (let's see what GitHub makes of it: perhaps I should just `M-x report-emacs-bug`). 

> I definitely consider this API as "not yet released".
> You mean we'd pass the callback function as an argument?
> That's a valid option as well, indeed.

In that case I would simplify my patch more and remove the `eldoc-documentation-callback` thing, and just pass the callback as an arg.

```diff
cat 0001-Better-handle-asynchronously-produced-eldoc-docstrin.patch                                                                                                                                                                                                                               master ⬆ ✭ ◼
From cd6ef1b12ea120773cbe239157f06eca93a932cd Mon Sep 17 00:00:00 2001
From: =?UTF-8?q?Jo=C3=A3o=20T=C3=A1vora?= <joaotavora@gmail.com>
Date: Mon, 25 May 2020 16:39:40 +0100
Subject: [PATCH] Better handle asynchronously produced eldoc docstrings

* lisp/emacs-lisp/eldoc.el (eldoc-documentation-functions):
Overhaul docstring.
(eldoc-documentation-compose, eldoc-documentation-default): Handle
non-nil, non-string values of elements of
eldoc-documentation-functions.  Use eldoc--handle-multiline.
(eldoc-print-current-symbol-info): Honour non-nil, non-string
values returned by eldoc-documentation-callback.
(eldoc--handle-multiline): New helper.
(Version): Bump to 1.1.0.
---
 lisp/emacs-lisp/eldoc.el | 73 ++++++++++++++++++++++++++++++----------
 1 file changed, 56 insertions(+), 17 deletions(-)

diff --git a/lisp/emacs-lisp/eldoc.el b/lisp/emacs-lisp/eldoc.el
index ef5dbf8103..288f617822 100644
--- a/lisp/emacs-lisp/eldoc.el
+++ b/lisp/emacs-lisp/eldoc.el
@@ -5,7 +5,7 @@
 ;; Author: Noah Friedman <friedman@splode.com>
 ;; Keywords: extensions
 ;; Created: 1995-10-06
-;; Version: 1.0.0
+;; Version: 1.1.0
 ;; Package-Requires: ((emacs "26.3"))
 
 ;; This is a GNU ELPA :core package.  Avoid functionality that is not
@@ -338,12 +338,26 @@ eldoc-display-message-no-interference-p
 
 

 (defvar eldoc-documentation-functions nil
-  "Hook for functions to call to return doc string.
-Each function should accept no arguments and return a one-line
-string for displaying doc about a function etc. appropriate to
-the context around point.  It should return nil if there's no doc
-appropriate for the context.  Typically doc is returned if point
-is on a function-like name or in its arg list.
+  "Hook of functions that produce doc strings.
+Each hook function should accept no arguments and decide whether
+to display a doc short string about the context around point.  If
+the decision and the doc string can be produced quickly, the hook
+function should immediately return the doc string, or nil if
+there's no doc appropriate for the context.  Otherwise, if its
+computation is expensive or can't be performed directly, the hook
+function should save the value bound to
+`eldoc-documentation-callback', and arrange for that callback
+function to be asynchronously called at a later time, passing it
+either nil or the desired doc string.  The hook function should
+then return a non-nil, non-string value.
+
+A current limitation of the asynchronous case is that it is only
+guaranteed to work correctly if the value of
+`eldoc-documentation-function' (notice the singular) is any other
+than `eldoc-documentation-default'.
+
+Typically doc is returned if point is on a function-like name or
+in its arg list.
 
 Major modes should modify this hook locally, for example:
   (add-hook \\='eldoc-documentation-functions #\\='foo-mode-eldoc nil t)
@@ -351,14 +365,18 @@ eldoc-documentation-functions
 taken into account if the major mode specific function does not
 return any documentation.")
 
+(defun eldoc--handle-multiline (res)
+  "Helper for handling a bit of `eldoc-echo-area-use-multiline-p'."
+  (if eldoc-echo-area-use-multiline-p res
+    (truncate-string-to-width
+     res (1- (window-width (minibuffer-window))))))
+
 (defun eldoc-documentation-default ()
   "Show first doc string for item at point.
 Default value for `eldoc-documentation-function'."
   (let ((res (run-hook-with-args-until-success 'eldoc-documentation-functions)))
-    (when res
-      (if eldoc-echo-area-use-multiline-p res
-        (truncate-string-to-width
-         res (1- (window-width (minibuffer-window))))))))
+    (cond ((stringp res) (eldoc--handle-multiline res))
+          (t res))))
 
 (defun eldoc-documentation-compose ()
   "Show multiple doc string results at once.
@@ -368,13 +386,11 @@ eldoc-documentation-compose
      'eldoc-documentation-functions
      (lambda (f)
        (let ((str (funcall f)))
-         (when str (push str res))
+         (when (stringp str) (push str res))
          nil)))
     (when res
       (setq res (mapconcat #'identity (nreverse res) ", "))
-      (if eldoc-echo-area-use-multiline-p res
-        (truncate-string-to-width
-         res (1- (window-width (minibuffer-window))))))))
+      (eldoc--handle-multiline res))))
 
 (defcustom eldoc-documentation-function #'eldoc-documentation-default
   "Function to call to return doc string.
@@ -408,6 +424,12 @@ eldoc--supported-p
            ;; there's some eldoc support in the current buffer.
            (local-variable-p 'eldoc-documentation-function))))
 
+;; this variable should be unbound, but that confuses
+;; `describe-symbol' for some reason.
+(defvar eldoc-documentation-callback nil
+  "Dynamically bound.  Accessible to `eldoc-documentation-functions'.
+See that function for details.")
+
 (defun eldoc-print-current-symbol-info ()
   "Print the text produced by `eldoc-documentation-function'."
   ;; This is run from post-command-hook or some idle timer thing,
@@ -417,11 +439,28 @@ eldoc-print-current-symbol-info
         ;; Erase the last message if we won't display a new one.
         (when eldoc-last-message
           (eldoc-message nil))
-      (let ((non-essential t))
+      (let ((non-essential t)
+            (buffer (current-buffer)))
         ;; Only keep looking for the info as long as the user hasn't
         ;; requested our attention.  This also locally disables inhibit-quit.
         (while-no-input
-          (eldoc-message (funcall eldoc-documentation-function)))))))
+          (let*
+              ((waiting-for-callback nil)
+               (eldoc-documentation-callback
+                (lambda (string)
+                  (with-current-buffer buffer
+                    ;; JT@2020-05-25: Currently, we expect one single
+                    ;; docstring from the client, we silently swallow
+                    ;; anything the client unexpectedly gives us,
+                    ;; including updates.  This could change.
+                    (when waiting-for-callback
+                      (eldoc-message (eldoc--handle-multiline string))
+                      (setq waiting-for-callback nil)))))
+               (res
+                (funcall eldoc-documentation-function)))
+            (cond ((stringp res) (eldoc-message res))
+                  (res (setq waiting-for-callback t))
+                  (t (eldoc-message nil)))))))))
 
 ;; If the entire line cannot fit in the echo area, the symbol name may be
 ;; truncated or eliminated entirely from the output to make room for the
```

joaotavora:
The docstring contained an error: it should read: "is only guaranteed to work if the value is eldoc-documentation-default"

dgutov:
> In hindsight
it was bad to make elements eldoc-documentation-functions argless functions.

I disagree. The "futures" solution will most likely have argless functions. So the ad-hoc solution should use them too, then the API could be extended/swapper very easily.

> You could have followed a similar path to flymake-diagnostic-functions for
what is a completely analogous problem. (yes Dmitry, Flymake could also
use futures :-) ).

Naturally. There are other places they could be used in, too.

dgutov:
> I think what I like about futures is that it opens up the possibility to
extend the API a bit more easily.  E.g. adding the possibility to abort
a future.

Yup. Or add an "errback" as well, in which ElDoc will display/log in the execution error using some of its mechanisms.

joaotavora:
> I disagree. The "futures" solution will most likely have argless functions. So the ad-hoc solution should use them too, then the API could be extended/swapper very easily.

Well the future isn't here yet, is it? When it does come, it's a question of telling elements of `eldoc-documentation-functions` that they can ignore the `callback` arg and start returning a "future" object instead. You'll want a similar scheme for Flymake backends anyway (which has more users, I think).

Anyway, as much as I am convinced of the soundness of Stefan's proposal on futures (it is a reasonably easy concept, though i suspect a bit niche still), I would consider the possibility that we might be over-engineering it here, because I can't think of enormous  advantages of futures to Eldoc specifically:

1. aborting an eldoc call is mostly useless, you're saving a few moments of string manipulation before handing the string to Eldoc (and even those we're thinking of mostly offloading to Eldoc).

2. Eldoc doesn't have anything reasonably useful to do with a backend that reports that it tried to get a docstring and couldn't.

Curiously, applying the reasoning to Flymake yields opposite results:  Flymake does have a compelling use case for those two things.

1. Making diagnostics once you get them from the external process is reasonably expensive and O(N) with buffer size and in the middle of a potentially lengthy process filter. Being informed that the operation is going stale is a good thing. Best case: kill process immediately.

2. Flymake collects error reports from the backends, temporarily disabling them if they misbehave a lot.

dgutov:
> Well the future isn't here yet, is it?

I just showed an ad-hoc API that would work very close to it.

> When it does come, it's a question of telling elements of eldoc-documentation-functions that they can ignore the callback arg and start returning a "future" object instead.

Why do this ugly thing when we can do it better from the start?

> aborting an eldoc call is mostly useless, you're saving a few moments of string manipulation before handing the string to Eldoc (and even those we're thinking of mostly offloading to Eldoc)

It would most of the time abort an HTTP request, not some sort of computation.

joaotavora:
> Why do this ugly thing when we can do it better from the start?

It's not ugly at all.  Just as ugly as multiple semantics for return values.  And if you want to do the future thing now, be my guest. I await your implementation.

> I just showed an ad-hoc API that would work very close to it.

Go for it. But why go for an ad-hoc API when...

> It would most of the time abort an HTTP request, not local computation.

How do you abort those?  I.e. where you would be checking for the abort flag in the future object?

joaotavora:
Two more points: 

1. I can't quite understand your `company` "future" thing as well as I can understand Stefan's idea, which is in line with what I expect from futures.

2. You can write an `aborted-p` for callbacks as well, no futuristic technology needed.

dgutov:
> I await your implementation.

OK.

> But why go for an ad-hoc API when...

It's about equal to complexity of your patch. But if we can get the futures in sooner rather than later, that'd be better.

> How do you abort those? I.e. where you would be checking for the abort flag in the future object?

By calling `(future-abort fut)`. And the code that created the future will make sure that this will do something useful (or not).

> I can't quite understand your company "future" thing as well as I can understand Stefan's idea, which is in line with what I expect from futures.

You probably know where to look for its documentation, but OK.

Please wait for the alternative patch.

joaotavora:
> By calling (future-abort fut). 
> And the code that created the future will make sure that this will do something useful (or not).

And exactly _when_ will that code do that in the HTTP request scenario that you presented?

dgutov:
Perhaps Eldoc will do that in `pre-command-hook`? After all, it wouldn't do for the callback to fire right after the user has just typed their text character? Or did something else invalidating the result.

dgutov:
Or, with `while-no-input`, it can do that right after if `while-no-input` returned t.

joaotavora:
> Perhaps Eldoc will do that in pre-command-hook? After all, it wouldn't do for the callback to fire right after the user has just typed their text character? Or did something else invalidating the result.

That problem is already solved by `run-with-idle-timer`, which `eldoc.el` already uses.

> Or, with while-no-input, it can do that right after if while-no-input returned t.

By the time that could be useful the presumed HTTP request is already well on its way and you're waiting for a filter or a sentinel call.  In this scenario, the only place where is might make sense to check `(future-aborted-p fut)` (or `(callback-aborted-p callback)`, for that matter) is in the process filter. But in the Eldoc case, the HTTP response will be extremely short, so you're gaining very little.

> It's about equal to complexity of your patch.

I seriously doubt that, BTW. To "get the futures in", i.e. to get a futures library in Emacs, you have to, well, develop a futures library.  A good one, presumably. That's a new file, and a new section in manual.  Not to mention a few weeks of bikeshedding (at best) with comparisons to existing implementations such as https://github.com/skeeto/emacs-aio (which, by the way, looks very good).

To hold up this needed fix while we wait for that to resolve seems nonsensical (and, I suppose, a bit paradigmatic of how the reddit crowd sees emacs development).

I say: let's develop a good futures library and use it as much as we can. It'll be your new baby. But babies need to gestate, so let's not rush it. In the meantime, let's fix this actual problem that Eglot is facing.  The `eldoc-documentations-functions` API is pretty much "unreleased", as Stefan says, so we can always change it afterwards.  At worst we put a big disclaimer there like you put in `xref.el`.

dgutov:
You set up the callback inside `run-with-idle-timer`. But when does it fire?

> is in the process filter. But in the Eldoc case, the HTTP response will be extremely short, so you're gaining very little

Depending on the implementation, it could be parsing a huge JSON or whatever. And some other implementations could terminate the network connection (if they expect the server on the other side to handle it well), or kill an external process.

joaotavora:
> You set up the callback inside run-with-idle-timer. But when does it fire?

I don't understand. As soon as the response arrives, of course.  As soon as the idle timer fires, you'll have made the request, there's no stopping it.There's also no point in checking `while-no-input` there because the definition of idle timer is that input has stopped for a while.

> Depending on the implementation, it could be parsing a huge JSON or whatever. 

For eldoc? Nah. For diagnostics, maybe, indeed we get very large diagnostic responses in Eglot.

But if you really really think that aborting is a killer feature here, I can add a  `eldoc-callback-aborted-p` to my patch, it's pretty easy to implement.  Not sure if Eglot should bother checking it though.

> And some other implementations could terminate the network connection (if they expect the server on the other side to handle it well), or kill an external process.

Again, the only change they have to do so in the process filter or sentinel.  So by then you'll have wasted the effort.

dgutov:
>  As soon as the response arrives, of course.

And is Emacs idle at that time?

> There's also no point in checking while-no-input there because the definition of idle timer is that input has stopped for a while.

Yes, true. Only having done a very cursory look at the code, I imagined it doing a sit-for loop until the response arrives. I can see now this is not the case.

> For eldoc? Nah. For diagnostics, maybe, indeed we get very large diagnostic responses in Eglot.

This is not only for Eglot, though.

> Not sure if Eglot should bother checking it though.

No idea, really.

> Again, the only chance they have to do so in the process filter or sentinel.

Not sure why you think so.

Anyway, check out https://debbugs.gnu.org/41531#8.

joaotavora:
> And is Emacs idle at that time?

Doesn't matter, by the time you can check if input is pending, you'll have bothered the server and Emacs will have slurped a good chunk of its response into memory.

> This is not only for Eglot, though.

Yes, but what "function signature at point" server produces "huge JSON" then?

> > Again, the only chance they have to do so in the process filter or sentinel. So by then you'll have wasted the effort.
> Not sure why you think so.

Because that's the way Emacs works! The only effort you can save is heavy lisp processing in the process filter.  And what I'm telling you is that the effort is reasonably small _in the eldoc_ case.

> Anyway, check out https://debbugs.gnu.org/41531#8.

Thanks.  Doesn't look bad. For a poor man's async primitive, I prefer my version, which matches what flymake.el and url.el (and possibly also websocket.el) do.  But of course I can live with your patch, so I'll let Stefan or someone else decide .  Let's continue in the Emacs bug tracker.

dgutov:
> Doesn't matter, by the time you can check if input is pending, you'll have bothered the server and Emacs will have slurped a good chunk of its response into memory.

OK, another question: if the result still /valid/?

> Yes, but what "function signature at point" server produces "huge JSON" then?

No idea, a hypothetical one. It's an open API, not everybody is or will be using LSP until the end of time. And it doesn't really have to be huge. A saving is a saving.

> The only effort you can save is heavy lisp processing in the process filter.

You can certainly kill the external process outside of it. Saving on CPU expenses in general.

> For a poor man's async primitive, I prefer my version

So even the code savings didn't convince you? Both in eldoc.el, and likely in doc functions as well.

> which matches what flymake.el and url.el (and possibly also websocket.el) do

Using a global var is /so/ Emacs 23.

joaotavora:
Continued the rest of the conversation in the emacs bug tracker link you posted.

> Using a global var is so Emacs 23.

`eldoc--callback` isn't a global var, it's a special var. Cause Lisp is /special/ :sparkles:

PR #469, opened 2020-05-11
Fix #454: prefer local-help in eglot-help-at-point

### LINKED ISSUE #454: Hover vs. diagnostics (C-h .)
Commit a0249c8 remapped display-local-help to eglot-help-at-point.  I used `C-h .` to show diagnostics errors at point.  Now if there are documentation and diagnostics at point, `C-h .` shows the former.  Is there a keybinding to show the latter?  If not, would it be possible to prioritize the diagnostics info and only show the documentation when, for example, `C-h .` is invoked with a prefix argument?

Thanks.

nemethf:
Commit a0249c8 remapped display-local-help to eglot-help-at-point.
This commit restores the old behaviour when both diagnostics and hover
information are available, i.e., it shows the diagnostics information.

* eglot.el (eglot-help-at-point): Prioritize local-help over hover
info.

### DIFF
--- a/eglot.el
+++ b/eglot.el
@@ -2258,19 +2258,22 @@ is not active."
 (defun eglot-help-at-point ()
   "Request documentation for the thing at point."
   (interactive)
-  (eglot--dbind ((Hover) contents range)
-      (jsonrpc-request (eglot--current-server-or-lose) :textDocument/hover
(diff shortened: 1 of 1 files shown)


nemethf:
#454 has been fixed in a different way.  This PR is not needed anymore.

PR #477, opened 2020-05-13
add automatic detection for Fortran

tukss:
The Fortran language server `fortls` works well with eglot. I tested both `fortran-mode` for "old" Fortran and `f90-mode`.


skangas:
@joaotavora This LGTM, WDYT?

skangas:
Actually, we already added Fortran support in #603, so the only thing remaining here is to also use it for f90-mode. So I modified the commit to do that, and merged it to master.

@tukss Thank you for your contribution!

PR #481, opened 2020-05-16
Display flymake diagnostics along with documentation

dandavison:
I love the syntax-highlighted documentation provided by `eglot-help-at-point`. (I even bind it to `?` and use `C-u ?` for a literal question mark.) What I'm finding is that, in practice, when there is a compiler error at point, I'd like to see that error as part of the help output.

I understand that this may be mixing two sources of information -- one of which is flymake's concern. But is there any sympathy for this, in spirit at least? (the implementation here is just a quick proof-of-principle.) Or, am I missing a way to display syntax-highlighted diagnostics, or is there a different way this should be done?

<table>
  <thead>
    <td><emph>Before</emph></td>
    <td><emph>After</emph></td>
  </thead>
  <tr>
    <td>
      image
    </td>
    <td>
      image
    </td>
  </tr>
  <tr>
    <td>
      image
    </td>
    <td>
      image
    </td>
  </tr>
</table>

### DIFF
--- a/eglot.el
+++ b/eglot.el
@@ -2268,12 +2268,15 @@ is not active."
     (let ((blurb (and (not (seq-empty-p contents))
                       (eglot--hover-info contents range))))
       (if blurb
-          (with-current-buffer (eglot--help-buffer)
-            (with-help-window (current-buffer)
-              (rename-buffer (format "*eglot-help for %s*"
-                                     (thing-at-point 'symbol)))
-              (with-current-buffer standard-output (insert blurb))
-              (setq-local nobreak-char-display nil)))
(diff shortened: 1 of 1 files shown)


nemethf:
I think we started to discuss your idea in #454.  At any rate, your proposal can be generalized to show any local help and not just flymake diagnostics.

joaotavora:
Yes, this is a good idea, but please let's close this and continue in that
other issue. You can repost your screenshots there, if you wish.

João

joaotavora:
@dandavison, very soon this idea will be possible with the latest Flymake 1.0.9 and the newest Eglot changes.

PR #482, opened 2020-05-18
Use filter-buffer-substring to get buffer text

muffinmad:
Python file:

```python
int
```

`textDocument/hover` response:

```
(:id 44 :jsonrpc "2.0" :result
     (:contents
      (:kind "markdown" :value "```python\nint(x: Union[Text, bytes, SupportsInt, _SupportsIndex]=...)\nint(x: Union[Text, bytes, bytearray], base: int)\n```\n\n```\nint([x]) -> integer\nint(x, base=10) -> integer\n\nConvert a number or string to an integer, or return 0 if no arguments\nare given.  If x is a number, return x.__int__().  For floating point\nnumbers, this truncates towards zero.\n\nIf x is not a number or if base is given, then x must be a string,\nbytes, or bytearray instance representing an integer literal in the\ngiven base.  The literal can be preceded by '+' or '-' and be surrounded\nby whitespace.  The base defaults to 10.  Valid bases are 0 and 2-36.\nBase 0 means to interpret the base from the string as an integer literal.\n>>> int('0b100', base=0)\n4\n```")
      :range nil))
```

With default eglot settings the following message is shown in the echo area:
```
[eglot]
(...truncated. Full help is in `*eglot-help for int*')
```

Codeblock markers are invisible in `gfm-view-mode` and mostly useless after font faces are applied to the text.

This PR removes codeblock markers from the hover info.

### DIFF
--- a/eglot.el
+++ b/eglot.el
@@ -1162,10 +1162,10 @@ Doubles as an indicator of snippet support."
                          (_ major-mode))))))
     (with-temp-buffer
       (setq-local markdown-fontify-code-blocks-natively t)
-      (insert (string-trim string))
+      (insert string)
       (ignore-errors (delay-mode-hooks (funcall mode)))
       (font-lock-ensure)
-      (buffer-string))))
+      (string-trim (filter-buffer-substring (point-min) (point-max))))))
 
 (defcustom eglot-ignored-server-capabilites (list)
   "LSP server capabilities that Eglot could use, but won't.


eglot.el
```diff
@@ -1162,10 +1162,16 @@ Doubles as an indicator of snippet support."
                          (_ major-mode))))))
     (with-temp-buffer
       (setq-local markdown-fontify-code-blocks-natively t)
-      (insert (string-trim string))
+      (insert string)
       (ignore-errors (delay-mode-hooks (funcall mode)))
       (font-lock-ensure)
-      (buffer-string))))
+      (when (eq major-mode 'gfm-view-mode)
```
joaotavora, line 1168: This is too much for `eglot.el`, @muffinmad, sorry.  IT's clearly something that is to be done in the server, in `gfm-view-mode`, or maybe in a user hook.  But I think you should contribute this to `gfm-view-mode`.

muffinmad: I see. Would it be fine for `eglot` to use `filter-buffer-substring` instead of `buffer-string` here? In this case `gfm-view-mode` (among other modes) can provide own `filter-buffer-substring-function`.

joaotavora: I didn't know about that function, but that sounds very good in principle, yes.

muffinmad:
<s>This PR removes codeblock markers from the hover info.</s>

This PR allows major modes, used to format the hover info text, to filter the text.

joaotavora:
I think this PR now makes sense.  However, it is too soon to merge it, probably. I'll make the most sense when `gfm-view-mode` starts supporting this: have you contacted its authors?  Or are you planning to set your own filter from your `.emacs`? What would this filter look like?

muffinmad:
> I think this PR now makes sense. However, it is too soon to merge it, probably.

OK, no problem.

> I'll make the most sense when `gfm-view-mode` starts supporting this: have you contacted its authors?

Not yet.

> Or are you planning to set your own filter from your `.emacs`? What would this filter look like?

Something like this

```elisp
(defun test/filter (start end &optional delete)
  (replace-regexp-in-string "^```.*\n*?" ""
                            (buffer-substring start end)))

(add-hook 'gfm-view-mode-hook
          (lambda ()
            (setq-local filter-buffer-substring-function
                        #'test/filter)))
```

But this will not work because of `delay-mode-hooks`, right?

joaotavora:
> But this will not work because of delay-mode-hooks, right?

You lost me :-)

muffinmad:
> You lost me :-)

Sorry :) My local `.emacs` solution add the hook for `gfm-view-mode`.  That hook will set the `filter-buffer-substring-function` local variable to `test/filter` function that will be used to filter buffer text.  But that hook will not be called because `gfm-view-mode` is enabled with `delay-mode-hooks` in the `eglot--format-markup` function.

Here is another way to make it solely in `eglot` without dependency on third-party modules :-) E.g.

```elisp
(let ((filter-buffer-substring-function eglot-filter-help-buffer-function))
  (filter-buffer-substring (point-min) (point-max)))
```

Then the `eglot-filter-help-buffer-function` variable can be `'identity` by default and users can write their own filter functions.  Doesn't really like this solution bacouse in this case we need to add questionable user option to `eglot`.

muffinmad:
Too bad there are no something like `buffer-string-only-visible` function in Emacs.

joaotavora:
bah.  `vc-region-history` worked perfectly:
```
commit 2ab8b59db44fd4dd9791ac203ac9f0915cc6824f
Author: Xu Chunyang <4550353+xuchunyang@users.noreply.github.com>
Date:   Sun Oct 27 23:41:53 2019 +0800

    Don't run mode hooks in eglot--format-markup
    
    * eglot.el (eglot--format-markup): Use delay-mode-hooks.
    
    Copyright-paperwork-exempt: yes

--- a/eglot.el
+++ b/eglot.el
@@ -1086,1 +1086,1 @@
-      (ignore-errors (funcall mode)) (font-lock-ensure) (buffer-string))))
+      (ignore-errors (delay-mode-hooks (funcall mode)))
```

But smart old me forgot to include a reference to the issue that would have explained why `delay-mode-hooks` is there in the first place...

joaotavora:
Got it, it's #327. The `delay-mode-hooks` is there because of "normal", programming major modes, which normally contains looks of user hooks.  We could remove it for `gfm-view-mode`, and so your hook would work. But the correct version is to contact the authors of `gfm-view-mode` anyway, and if you really want a hook that works regardless of `delay-mode-hooks`, just use `add-function` and it will do the trick.

muffinmad:
I'm mostly use `vc-annotate` for such cases ;-)

Call of `delay-mode-hooks` function is in it's place. We don't need to run all those mode hooks to just apply font faces.

muffinmad:
> But the correct version is to contact the authors of `gfm-view-mode` anyway, and if you really want a hook that works regardless of `delay-mode-hooks`, just use `add-function` and it will do the trick.

Absolutely agree. Thanks for the tips!

joaotavora:
> I'm mostly use vc-annotate for such cases ;-)

Yeah, but that annotates the whole buffer.  I like `vc-region-history` much better, personally.

muffinmad:
> But the correct version is to contact the authors of `gfm-view-mode`

`filter-buffer-substring-function` is now implemented in both `markdown-view-mode` and `gfm-view-mode`. See https://github.com/jrblevin/markdown-mode/pull/494.

joaotavora:
> filter-buffer-substring-function is now implemented in both markdown-view-mode and gfm-view-mode

Nice, but what does that mean for us?

muffinmad:
> > filter-buffer-substring-function is now implemented in both markdown-view-mode and gfm-view-mode
> 
> Nice, but what does that mean for us?

That means to us that this PR must be merged immediately ;-) And make eglot respect `filter-buffer-substring-function` provided by the major mode used to present the hover info to happy eglot users!

joaotavora:
Andrii Kolomoiets <notifications@github.com> writes:

>  filter-buffer-substring-function is now implemented in both
>  markdown-view-mode and gfm-view-mode
>
>  Nice, but what does that mean for us?
>
> That means to us that this PR must be merged immediately ;-) And make
> eglot respect filter-buffer-substring-function provided by the major
> mode used to present the hover info to happy eglot users!

Doh.  Sorry, I got confused.  Merged.

PR #494, opened 2020-06-02
Don't update echo area when textDocument/hover returned nil

dgutov:
That creates blinking when it arrives before (non-empty) signatureHelp response.

### DIFF
--- a/eglot.el
+++ b/eglot.el
@@ -2401,11 +2401,11 @@ documentation.  Honour `eglot-put-doc-in-help-buffer',
         (jsonrpc-async-request
          server :textDocument/hover position-params
          :success-fn (eglot--lambda ((Hover) contents range)
-                       (unless sig-showing
+                       (unless (or sig-showing
+                                   (seq-empty-p contents))
                          (when-buffer-window
-                          (eglot--update-doc (and (not (seq-empty-p contents))
-                                                  (eglot--hover-info contents
-                                                                     range))
+                          (eglot--update-doc (eglot--hover-info contents
+                                                                range)
                                              thing-at-point))))
          :deferred :textDocument/hover))
       (when (eglot--server-capable :documentHighlightProvider)


joaotavora:
Thanks, looks good. Can we have GNU Changelog-style commit messages, just like in Emacs?

muffinmad:
This will break #439. 

@dgutov Can you please provide `emacs -Q` receipt for why this behavior is needed?

joaotavora:
Good catch Andrii. Yes, let's look  at this more carefully then.

Also

joaotavora:
Indeed , this is another indication that what's needed is some deep work in `eldoc.el`: it must decide on a protocol for clients to follow and take responsibility to show or hide things in echo area, instead of having the clients worry about that.  That should resolve the blinking _and_ the #439.  For now, we live with the blinking.

dgutov:
> it must decide on a protocol for clients to follow and take responsibility to show or hide things in echo area, instead of having the clients worry about that

What are you on about?

Eglot just shouldn't call `eldoc-message` (or more) twice per user command. You need perhaps some asynchronous abstraction, like futures/promises and some code to merge them.

dgutov:
Whatever we decide on regarding eldoc's asynchronous convention, your code will need to be fixed. It might as well be fixed now.

dgutov:
@muffinmad Do you still need an example? I think the situation is pretty clear, though. And I can trivially show an excerpts of the log with interaction with the server.

joaotavora:
> What are you on about?

I meant exactly what I wrote, and if you're not going to be constructive and courteous, we might as well stop the conversation here.

For your benefit, I explain again: eldoc must have a protocol whereby it asks the client "what, if any, is the documentation at this spot" and then gives the client a way to inform about that either immediately or at a later time.  And that client might want to inform once, twice, no times at all. Eldoc will then decide which of those to consider of the things to show. It must NOT give the client direct access to the echo area, which is what happens now.

> Whatever we decide on regarding eldoc's asynchronous convention, your code will need to be fixed. It might as well be fixed now.

Of course Eglot's code has to adapt to the new protocol. Until there's no such protocol, there's nothing to fix without breaking other more important stuff.  Except Eldoc, which, for the nth time,is proven to be insufficient for async clients.

> You need perhaps some asynchronous abstraction, like futures/promises and some code to merge them.

Exactly, except your personal preference for futures/promises is absolutely orthogonal to this, and foreign to the way Emacs solves these problems.

dgutov:
To retain the spirit of courtesy, maybe avoid closing unfinished discussions?

> eldoc must have a protocol whereby it asks the client "what, if any, is the documentation at this spot" and then gives the client a way to inform about that either immediately or at a later time. And that client might want to inform once, twice, no times at all.

Err, no. Neither of the proposals, unless I'm really missing something, will allow the client to send back multiple responses. At best, all but the first one will be ignored. The result, in the scenario that I tried to fix here, will be an (almost always) empty echo area.

So the responsibility of merging the responses will stay on the client, not in the Eldoc code.

> It must NOT give the client direct access to the echo area, which is what happens now.

That is true. But try to imagine the change to Eglot you would do if your callback proposal was accepted. Can you show the patch you would apply in such a case?

> Exactly, except your personal preference for futures/promises is absolutely orthogonal to this, and foreign to the way Emacs solves these problems.

How you are going to solve this internally is indeed not my call. A utility function like `aio-all` could help, though.

muffinmad:
> @muffinmad Do you still need an example? I think the situation is pretty clear, though. And I can trivially show an excerpt of the log with interaction with the server.

Sure. It's always good to have the receipt to reproduce the issue starting from `emacs -Q`, isn't it?

Eglot doesn't blink for me, so the issue may depend on the lsp server used, some third-party package etc.

dgutov:
The scenario is like this:

1. Install `elixir-ls`.
2. Launch eglot in a buffer in `elixir-mode` with contents like `Enum.chunk([], 2)`.
3. Move around inside parens (`C-f`, `C-b`, repeat).

The echo area will blink more than it has to.

The log shows:

```
[client-request] (id:88) Tue Jun  2 16:40:09 2020:
(:jsonrpc "2.0" :id 88 :method "textDocument/signatureHelp" :params
          (:textDocument
           (:uri "file:///home/dgutov/xyz.ex")
           :position
           (:line 25 :character 18)))
[client-request] (id:89) Tue Jun  2 16:40:09 2020:
(:jsonrpc "2.0" :id 89 :method "textDocument/hover" :params
          (:textDocument
           (:uri "file:///home/dgutov/xyz.ex")
           :position
           (:line 25 :character 18)))
[server-reply] (id:89) Tue Jun  2 16:40:09 2020:
(:id 89 :jsonrpc "2.0" :result nil)
[server-reply] (id:88) Tue Jun  2 16:40:09 2020:
(:id 88 :jsonrpc "2.0" :result
     (:activeParameter 1 :activeSignature 0 :signatures
                       [(:label "chunk(enumerable, count)" :parameters
                                [(:label "enumerable")
                                 (:label "count")])
                        (:label "chunk(enum, n, step)" :parameters
                                [(:label "enum")
                                 (:label "n")
                                 (:label "step")])
                        (:label "chunk(enumerable, count, step, leftover)" :parameters
                                [(:label "enumerable")
                                 (:label "count")
                                 (:label "step")
                                 (:label "leftover")])]))
```

The crux of the problem: `hover` sends its response faster than `signatureHelp`.

joaotavora:
> To retain the spirit of courtesy, maybe avoid closing unfinished discussions?

Closing the issue doesn't close the discussion, just signals my intention to not patch this like this. That may change. So just keep discussing and move on.  I hope I don't have to lecture on why "what are you on about" isn't the right way to express confusion in a discussion.

> That is true. But try to imagine the change to Eglot you would do if your callback proposal was accepted. Can you show the patch you would apply in such a case?

Eglot would probably have two sources of eldoc, two members of `eldoc-documentation-functions`. That's the cleanest way.  Another option is for eldoc to ignore the second thing that a specific client feeds it for a specific eldoc request.  Or update with the latest.  But all this would require you to stop holding up the patch because of a fixation with a specific async implementation.

> How you are going to solve this internally is indeed not my call. A utility function like aio-all could help, though.

Sure, but we don't need that stuff to fix this, as I've shown countless times.  I'm not going to start depending on a futures just to fix this. Please stop pushing a fetish async abstraction: it's completely orthogonal to this problem.

dgutov:
> Please stop pushing a fetish async abstraction

Speaking of courtesy, could you start speaking normally? This is disrespectful.

> Eglot would probably have two sources of eldoc, two members of eldoc-documentation-functions. That's the cleanest way.

Something to be considered. But that would break the existing semantics where some modes expect their result to take priority based on the position inside the hook. Not on when the first response arrives.

Likewise, it would run counter to the semantics of the current usages where the modes used `advice-add` with a specific ordering, and then were sure about the resulting priority.

> But all this would require you to stop holding up the patch because of a fixation with a specific async implementation.

You could try to start collaborating and use the requested abstraction, though. Then we can move on to other concerns, and the eldoc-specific futures could be swapped for a "global" implementation. Or, okay, let's wait until the proposal for the latter is ready.

joaotavora:
> Speaking of courtesy, could you start speaking normally? This is disrespectful.

I hereby apologize.

>The crux of the problem: hover sends its response faster than signatureHelp does.

I disagree, the "crux" of the problem is that eldoc doesn't serve Eglot's needs.  So until it does, it's better to live with this excessive blinking and not break #439.  Which is why I'm keeping this issue closed.

> Something to be considered. But that would break the existing semantics where some modes expect their result to take priority based on the position inside the hook

No, that just depends on the setting of `eldoc-documentaiton-function` (_singular_).  Eglot would set that as well, to `eldoc-documentation-compose` as it already does to a lot of other functions. Of course major modes can ask Eglot to stay out of that and live with the flak. Earlier I said `eldoc-documentation-compose` is too far-fetched.  I was mistaken, it is not, not at all.

> You could try to start collaborating and use the requested abstraction, though. 

But why? Why are you "requesting" it? I don't oppose that you create other software inside Emacs with that abstraction, but why are you forcing it for this specific case? I don't want to push complex futures on Eglot developers. They're much harder to debug with edebug, for one.  But if _you_ want to use futures in your programs, maybe even your eldoc.el-using programs, I have _no_ objection to that.  Zero, zilch, nada.

dgutov:
> I disagree, the "crux" of the problem is that eldoc doesn't serve Eglot's needs.

:roll_eyes:

Okay, I see. You really want to offload that responsibility.

> Eglot would set that as well, to eldoc-documentation-compose as it already does to a lot of other functions.

Why? You just said in the other thread that it's a user preference.

> Of course major modes can ask Eglot to stay out of that and live with the flak.

Major modes???

> They're much harder to debug with edebug, for one.

What makes you say that?

joaotavora:
> Okay, I see. You really want to offload that responsibility.

Indeed I do, very much.  This is not a responsibility of Eglot. It fits much better in Eldoc, especially its its brand-new `functions` abstractions.  It just has to do async correctly.  

Offloading responsibility in programming is not laziness, it's design.  And it could not be laziness anyway, because I'm the one doing the work for eldoc to handle async correctly anyway.
  
> Major modes???

Yes.  Major modes can do that. As can users.  But you seem to be diverting to scenarios that don't exist and criticizing abstractions that already exist in `eldoc.el`.  So it's out of place in this issue.

> What makes you say that?

When you have a function that returns a function that takes a function, it's harder to follow edebug around.  It's easier when you have a function that takes a function. Then you decide which one to instrument. Same for `trace-function`.  It's just simpler.  Again, for me.  If you have some technique for debugging futures, `aio`-futures, company-style-futures, Stefan-style-futures, whatever, I'm not against it. But don't push it on me, please.

dgutov:
> Major modes can do that.

Major modes can't override minor modes' decisions. Minor modes run later.

> criticizing abstractions that already exist in eldoc.el

Criticizing the semantics you want to add? Maybe.

> When you have a function that returns a function that takes a function

With the futures (or promises) it will be: a function that returns a container (created with `(future-make)`, or whatever). And upon completion it passes the result value to that container with `(future-resolve fut value)`. Something, somewhere (if not eldoc functions, then eldoc inernals), would have to create values that represent the unfinished computations anyway. And if such values are not reified (don't have proper names and thus legible presence in the call stack), debugging the whole process is likely to be harder, not easier.

joaotavora:
> Major modes can't override minor modes' decisions. Minor modes run later.

That's wrong: there's `eglot-stay-out-of` for that.  if you don't like the negative sound to it, it could be inverted to something else. 

But again, you're going out of your way to describe scenarios that don't exist yet.  Eglot is not yet in Emacs core, and for now works as a take-over-everything-type-extension.  Many people like that.

> Criticizing the semantics you want to add? Maybe.

No. dont' do that.  It's exhausting.  You're criticizing the relationship between `eldoc-documentation-function` and `eldoc-documentation-functions` . I didn't write that.  I just didn't.  I'm just proposing async support so that Eglot can use the existing semantics, which I merely opine that would be fine.

> debugging the whole process is likely to be harder, not easier.

Sure, if your brain likes to work like that, who am I to contradict it?  Work like that, by all means. But don't push that debugging strategies to others.

There is nothing more for this thread to give.

Issue #525, opened 2020-08-21
Wrong language ID for OCaml with Tuareg


arturovm:
Hi! I'm trying to use Eglot with Tuareg in OCaml files, and [ocaml-lsp](https://github.com/ocaml/ocaml-lsp). From what I can read in Eglot's source, it infers the language ID from the major mode's name and, since Tuareg's major mode name is `tuareg-mode`, it sends `tuareg` to the LSP server, which errors out as a result.

As far as I can tell, this is an edge case, but seeing that Tuareg seems to be the most widespread OCaml mode, and ocaml-lsp seems to be the "official" LSP implementation for OCaml, I suppose it would be worth tackling.

Do you have any pointers on how to solve this? I tried with the `.dir-locals.el` approach, but realized that that's only for the initial server config.

Here's the full event log, just in case:

```
client-request (id:1) Fri Aug 21 10:13:20 2020:
(:jsonrpc "2.0" :id 1 :method "initialize" :params
	  (:processId 63950 :rootPath "/[redacted]/OCaml/" :rootUri "file:///[redacted]/OCaml/" :initializationOptions nil :capabilities
		      (:workspace
		       (:applyEdit t :executeCommand
				   (:dynamicRegistration :json-false)
				   :workspaceEdit
				   (:documentChanges :json-false)
				   :didChangeWatchedFiles
				   (:dynamicRegistration t)
				   :symbol
				   (:dynamicRegistration :json-false)
				   :configuration t)
		       :textDocument
		       (:synchronization
			(:dynamicRegistration :json-false :willSave t :willSaveWaitUntil t :didSave t)
			:completion
			(:dynamicRegistration :json-false :completionItem
					      (:snippetSupport :json-false)
					      :contextSupport t)
			:hover
			(:dynamicRegistration :json-false :contentFormat
					      ["markdown" "plaintext"])
			:signatureHelp
			(:dynamicRegistration :json-false :signatureInformation
					      (:parameterInformation
					       (:labelOffsetSupport t)))
			:references
			(:dynamicRegistration :json-false)
			:definition
			(:dynamicRegistration :json-false)
			:declaration
			(:dynamicRegistration :json-false)
			:implementation
			(:dynamicRegistration :json-false)
			:typeDefinition
			(:dynamicRegistration :json-false)
			:documentSymbol
			(:dynamicRegistration :json-false :hierarchicalDocumentSymbolSupport t :symbolKind
					      (:valueSet
[...]

PR #542, opened 2020-09-22
Added the symbol attribute 'eglot-language-id

rvs314:
This implements the fix proposed by https://github.com/joaotavora/eglot/issues/525#issuecomment-678540557 to the issue of languages whose major mode has a different name from the language themselves

### DIFF
--- a/eglot.el
+++ b/eglot.el
@@ -1670,9 +1670,11 @@ THINGS are either registrations or unregisterations (sic)."
   (append
    (eglot--VersionedTextDocumentIdentifier)
    (list :languageId
-         (if (string-match "\\(.*\\)-mode" (symbol-name major-mode))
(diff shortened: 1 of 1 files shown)


rvs314:
I don't think the tests failing is a result of this patch - it seems like the failing tests are also failing on master. Running it on my machine, I get a subset of the errors from Travis (I imagine it's because I don't have all of the servers installed) with or without the patch.

joaotavora:
Yeah sorry to be so late with this. The test failures aren't your fault,
it's since Travis mix-up that I've been to busy to look into. I'll try to
merge your PR today.

João

purcell:
I have the same issue with `ocamllsp` and would love it if this could be merged.

PR #555, opened 2020-10-17
Add more information to the description of .dir-locals.el.

brown:
I added a description of how Emacs can be told to stop asking about the value that eglot-workspace-configuration is set to in a .dir-locals.el file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,12 +120,12 @@ file named `.dir-locals.el` in the project's root:
 This tells Emacs that any `python-mode` buffers in that directory
 should have a particular buffer-local value of
 `eglot-workspace-configuration`.  That variable's value should be
-_association list_ of _parameter sections_ which are presumably
+an _association list_ of _parameter sections_, which are presumably
 understood by the server.  In this example, we associate section
 `pyls` with the parameters object `(:plugins (:jedi_completion
 (:include_params t)))`.
 
-Now, supposing that you also had some Go code in the very same
+Now, supposing that you also have some Go code in the very same
 project, you can configure the Gopls server in the same file.  Adding
 a section for `go-mode`, the file's contents become:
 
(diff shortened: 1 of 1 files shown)


joaotavora:
Hello @brown , and sorry for the delay in answering this.  I find in `eglot.el`:

```
(put 'eglot-workspace-configuration 'safe-local-variable 'listp)
```

Which would mean that if `eglot-workspace-configuration` is a list, which it should apparently always be, there's no need for your README.md hint.  What are you setting `eglot-workspace-configuration` to in your `.dir-locals.el`?

brown:
Thanks very much for responding to my issue.

Your Emacs must be loading eglot.el before it loads the .dir-locals.el
file.  That's not the case for me.  I typically edit Java source files
in a directory tree that contains a .dir-locals.el file, then only
later decide to turn on eglot-mode.

joaotavora:
I see.  Then it makes sense, yes.  Well I don't have an answer handy except that I think this is Emacs's fault :-).  Or maybe if I put an autoload cookie on that expression it will do the trick.  In other words, this patch.
```
index 9944b18..1acc6fa 100644
--- a/eglot.el
+++ b/eglot.el
@@ -1803,6 +1803,7 @@ Records BEG, END and PRE-CHANGE-LENGTH locally."
 SECTION should be a keyword or a string, value can be anything
 that can be converted to JSON.")
 
+;;;###autoload
 (put 'eglot-workspace-configuration 'safe-local-variable 'listp)
 
 (defun eglot-signal-didChangeConfiguration (server)
```

Not sure how familar you are with autoloads and how you could test this.  But it should work.  I'll think a bit about this.

joaotavora:
> Not sure how familar you are with autoloads and how you could test this. But it should work. I'll think a bit about this.

Or maybe we can bother @monnier :-) Stefan this one should be easy. Thanks

monnier:
>> Not sure how familar you are with autoloads and how you could test
> this. But it should work. I'll think a bit about this.
> Or maybe we can bother @monnier :-) Stefan this one should be easy. Thanks

The autoload cookie looks right, yes.

PR #695, opened 2021-05-14
add ability to spread the notion of "current server" through xref

cmm:
* eglot.el (eglot-spread-through-xref): new defcustom, defaulting to
nil.
(eglot--xref-location-servers): new hash table, mapping file names to
servers.
(eglot--managed-mode): use eglot-current-server, instead of
eglot--cached-server directly.
(eglot--current-server-or-lose): ditto
(eglot--maybe-activate-editing-mode): ditto
(eglot-current-server): take xref in account
(eglot--xref-make-match): record the xref location

### DIFF
--- a/eglot.el
+++ b/eglot.el
@@ -237,6 +237,14 @@ let the buffer grow forever."
   :type '(choice (const :tag "Don't show confirmation prompt" nil)
                  (symbol :tag "Show confirmation prompt" 'confirm)))
 
+(defcustom eglot-spread-through-xref nil
+  "If non-nil, put anything reached through xref under originating server.
+This is handy if you tend to have one \"global\" or \"current\"
+project at any given time, and so don't mind nominally-common
+things (like system-installed libraries or header files) being
+considered part of one specific project."
+  :type 'boolean)
+
 ;; Customizable via `completion-category-overrides'.
 (when (assoc 'flex completion-styles-alist)
   (add-to-list 'completion-category-defaults '(eglot (styles flex basic))))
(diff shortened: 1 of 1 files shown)


eglot.el
```diff
@@ -1533,11 +1544,15 @@ Use `eglot-managed-p' to determine if current buffer is managed.")
 
 (defun eglot-current-server ()
   "Return logical EGLOT server for current buffer, nil if none."
-  eglot--cached-server)
+  (or eglot--cached-server
+      (setq eglot--cached-server
```
joaotavora, line 1548: This lazy "find the eglot--cached-server" is a good idea, especially, since you changed the previous references to use this accessor.

Do you see any problem in bringing all the eglot--cached-server calculation here, including the one in `eglot--maybe-activate-editing-mode`.  It seems even cleaner.

joaotavora (commented):
I like this one better than the other one.

joaotavora:
Much better! and I think you have a copyright assignment, so I'll have a tweak or two and merge over the weekend, maybe before if I find the time.

cmm:
(Just a local braindump, nothing actionable)

One (non-critical) deficiency of this implementation is its unfriendliness to `desktop-save` -- it would probably be better to map xref location to project roots instead of server objects, so that `eglot--servers-by-xref-location` could be added to `desktop-globals-to-save`.

But that would add at least one more problem on its own -- since eglot at present keys servers by the tuple [project-root, major-mode], it would break the spread-through-xref functionality when jumping from, say, a C++ source to a common C header.  I had a half-baked change somewhere that transitions the major-mode part of the key tuple to a predicate, can resurrect it if you are interested.  In any case ending up running 2 instances of ccls/clangd in the same project (one for C, the other for C++) is kind of sad...

But the bigger problem is that eglot is pretty much unusable with `desktop-save`/`desktop-restore` in general.  And I don't even know how to debug this (such magic, much hooks, very special startup behavior), so had to disable desktop mode altogether.

joaotavora:
Braindump much appreciated, but open a new discussion to handle this. In
it, you may safely assume that this pull request will be merged in this
current state.

In particular, I need to understand what desktop.el does and what it is not
doing correctly in the presence of eglot.

João

PR #700, opened 2021-05-29
Per #662, add a note to the README about updating TRAMP to 2.5.0.4+

jimporter:
This is just a small addition to the README so that people using TRAMP know that they should update if they're having problems. I also noticed that the `"reporting bugs"` anchor was in the wrong spot, so I moved that to the right place.


skangas:
Now merged, thanks for the contribution!

PR #702, opened 2021-06-03
Use eglot--executable-find to find remote executable

ilupin:


jimporter:
@joaotavora: To help explain what this PR is for, I filled out the issue template with all the relevant information. Since `eglot-alternatives` uses `executable-find` without the `remote` flag, it tries to find the LSP server executable on the local machine, even for TRAMP files. That makes Eglot unable to pick the right alternative over TRAMP, which is most noticeable when the alternative you want is *not* the first option.

---

(A side issue here is that`eglot-alternatives` uses `executable-find` to pick the alternative, and then I *think* the chosen alternative is passed to `executable-find` again in `eglot--guess-contact`: https://github.com/joaotavora/eglot/blob/e498cb171bb07ec36880a2494aafc8acb1cc34ca/eglot.el#L835

That second call could be eliminated if `eglot-alternatives` told us that the path has already been checked. This would speed up startup a bit, especially over TRAMP where `executable-find` takes a bit more work.)

---


     
* Server used: pylsp              
* Emacs version: GNU Emacs 27.2 (build 1, x86_64-w64-mingw32) of 2021-03-26            
* Operating system: MS Windows         
* Eglot version: 20210526.1751            
* Eglot installation method: MELPA/use-package      
* Using Doom: No               

#### LSP transcript - M-x eglot-events-buffer (mandatory unless Emacs inoperable)
N/A; this occurs before the LSP server starts.
    
#### Backtrace (mandatory, unless no error message seen or heard):
No error
   
#### Minimal configuration (mandatory)


First, install `python-lsp-server` on the remote machine. Then:
     
 ```sh
 # Type this in a shell to start an Emacs with Eglot configured
 $ /path/to/a/certain/version/of/emacs -Q -f package-initialize -L /path/to/git-cloned/eglot -l eglot.el 
 C-x C-f /ssh:server:/path/to/python-file.py
 M-x eglot
 ```
 
At this point, Eglot prompts the user with:
```
[eglot] I guess you want to run `pyls', but I can't find `pyls' in PATH!
Enter program to execute (or <host>:<port>): pyls
```
even though `pylsp` is installed and working on the remote system.

Another way to see this issue is to do the following in a remote buffer (i.e. where `default-directory` is a TRAMP path): `M-: (funcall  (eglot-alternatives '("pyls" "pylsp")))`. If `pylsp` is installed (and `pyls` isn't), you'd expect the result to be `("pylsp")`, but it's not.

joaotavora:
> A side issue here is thateglot-alternatives uses executable-find to pick the alternative, and then I think the chosen alternative is passed to executable-find again in eglot--guess-contact: 

If this is a side issue, make another issue for it.  I'm inclined to just merge this PR as is.

joaotavora:
@jimporter , for the side issue, can you check if this if this patch makes things better?  

```
index a6c60e9..c15e109 100644
--- a/eglot.el
+++ b/eglot.el
@@ -833,7 +833,9 @@ be guessed."
                      ((null guess)
                               managed-mode base-prompt))
-                     ((and program (not (eglot--executable-find program t)))
+                     ((and program 
+                           (not (file-name-absolute-p program))
+                           (not (eglot--executable-find program t)))
                       (concat (format "[eglot] I guess you want to run `%s'"
                                       program-guess)
                               (format ", but I can't find `%s' in PATH!" program)
```

jimporter:
> @jimporter , for the side issue, can you check if this if this patch makes things better?

Just on a quick inspection, I don't think that'll work as-is. The lambda from `eglot-alternatives` doesn't return the result of `eglot--executable-find`, so `program` wouldn't be absolute here. That is, if I call `(funcall (eglot-alternatives '("pyls" "pylsp")))`, the result is `'("pylsp")`, not `'("/path/to/pylsp")`. We could fix that though.

I'll get the rest of the info together and file a proper issue with all the details for this later today (hopefully).

jimporter:
@joaotavora Ok, I filed the side issue at #703. As mentioned there, it's a very minor issue and I'd be ok with you just closing the bug. I thought it was worth at least a mention though, since I noticed it while filling out the issue template for this PR.

PR #759, opened 2021-11-16
Fix issues #755, #401; severity not set in textDocument/publishDiagnostics

stephe-ada-guru:
Commits patch from #755


joaotavora:
Thanks @stephe-ada-guru .

Next time, you may press the arrow besides the green button, select "Rebase and merge", and press that instead.  This will avoid the introduction of Git spaghetti.

image

Anyway, minor issue and thanks for fixing this

PR #774, opened 2021-12-26
Use `locate-user-emacs-file` instead of `concat`

lorniu:
Not so hardcode, and any advices on `locate-user-emacs-file` will work.

### DIFF
--- a/eglot.el
+++ b/eglot.el
@@ -3013,8 +3013,8 @@ If INTERACTIVE, prompt user for details."
               (t "config_linux"))))
            (workspace
             (expand-file-name (md5 (project-root (eglot--current-project)))
-                              (concat user-emacs-directory
-                                      "eglot-eclipse-jdt-cache"))))
+                              (locate-user-emacs-file
+                               "eglot-eclipse-jdt-cache"))))
       (unless jar
         (setq jar
               (cl-find-if #'is-the-jar


skangas:
LGTM, it would just need a GNU-style ChangeLog and a marker "Copyright-paperwork-exempt: yes" at the end.

joaotavora:
Hi Stefan, in these cases, if the user is not familiar with these
conventions I just do it directly myself, either manually after fetching
the PR or directly from the GitHub web interface. It's fine to ask the
developer to do this himself/herself though.

skangas:
> in these cases, if the user is not familiar with these conventions I just do it directly myself, either manually after fetching the PR or directly from the GitHub web interface. It's fine to ask the developer to do this himself/herself though.

Sure, that's the plan. I'm still trying to figure out how to do that in a repository that I don't own myself. It doesn't seem very well documented, unfortunately.

joaotavora:
I think, having push rights, you can do anything from the command line.
I'll look into giving you more admin power :) or documenting this process
somehow.

lorniu:
Maybe like this? I'm not very familiar with this.

skangas:
> I think, having push rights, you can do anything from the command line. I'll look into giving you more admin power :) or documenting this process somehow.

It seems like I can just push directly to the repository as if I was the owner. So I will make an attempt, and we will see how much I break things. Hopefully not at all! ;-)

@lorniu Thanks!

skangas:
OK, it seems like that worked without a hitch!

Thanks for the contribution @lorniu and congratulations on your first commit to eglot!

joaotavora:
> OK, it seems like that worked without a hitch!

Yes, perfectly. Thanks very much! And you did well to rebase, not merge. If I should have added something it would be just to prefix `Fix #774:<space>` to the subject line in the commit message, but that's not strictly necessary at all.

@skangas there is also a NEWS.md file, but I don't necessarily enter all changes into it.  And it doesn't have to be updated in the same commit/PR.

PR #821, opened 2022-01-23
Properly check the CompletionItem.deprecated property

leungbk:
* eglot.el (eglot-completion-at-point): Check the :deprecated property
is `t'.  We do this so that a :deprecated property of :json-false does
not cause a completion candidate to be incorrectly marked as deprecated.

Verified with the Elixir language server with the following file:

``` elixir
IO.puts "Hello world!"
# Try typing the following:
# H

# The above should cause two deprecated completions, HashSet and HashDict, to be offered

# Also try typing the following:
# e

# The above should cause two non-deprecated completions.
```


skangas:
LGTM, thank you! Merged to master.

PR #838, opened 2022-02-21
Add lua-language-server to eglot-server-programs

leungbk:
* eglot.el (eglot-server-programs): Add lua-language-server. It has,
at the time of writing, more actively developed and more popular than
the lua-lsp server, so prioritize that.
* README.md: Advertise.


leungbk:

PR #855, opened 2022-02-28
Support window/showDocument request

nemethf:
Please review this carefully, because I haven't found any servers that send this request.  So I tested this against my own server-side implementation.  

(My plan is to modify rust-analyzer to use the standardized showDocument instead of their non-standard gotoLocation and I'd like if the new rust-analyzer feature would work in Eglot.) 

https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#window_showDocument

---

* eglot.el (eglot-client-capabilities): Add showDocument support.
(eglot-handle-request window/showDocument): New cl-defmethod.

### DIFF
--- a/eglot.el
+++ b/eglot.el
@@ -715,6 +715,7 @@ treated as in `eglot-dbind'."
                                        `(:valueSet
                                          [,@(mapcar
                                              #'car eglot--tag-faces)])))
+            :window '(:showDocument (:support t))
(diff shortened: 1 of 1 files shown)


nemethf:
Eglot has something like this since emacs-mirror/emacs@c083fa5cf80f711ac43dca1b6582aff1ad526e8e

Issue #847, opened 2022-02-25
Code actions missing (?) in javascript project


minikN:
Hello,

If this issue has been reported already I'm sorry, but I wasn't able to find any information on the issue tracker or Google in general.
I'd like to use eglot with react using the typescript-language-server. My setup is working, meaning I get diagnostics and can use code actions and so on.

For testing purposes, I created a new react app using create-react-app and ran npm install. Then I opened `src/App.js` (and modified it a bit):

```javascript
import React, { useContext } from 'react';
import logo from './logo.svg';
import './App.css';

function App() {
  const [test, setTest] = useState(0);
  return (
    <div className="App">
      <header className="App-header">
        <img src={logo} className="App-logo" alt="logo" />
        <p>
          Edit <code>src/App.js</code> and save to reload.
        </p>
        <a
          className="App-link"
          href="https://reactjs.org"
          target="_blank"
          rel="noopener noreferrer"
        >
          Learn React
        </a>
      </header>
    </div>
  );
}

export default App;
```

There are two things I don't understand how to use:

1) **AddMissingImports**: According to their doc, the typescript-language-server [supports](https://github.com/typescript-language-server/typescript-language-server#code-actions-on-save) adding missing imports (this also works using VSCode for example). However I don't know how to trigger it. In the example above, `useState` needs to be imported. But the action is not available when doing `eglot-code-actions`. I tried placing the cursor on the `u` of `useState`, select the word and select the line. It always reports `eglot--error: [eglot] No code actions here`. Is it simply not implemented yet or am I missing something?

2) **OrganizeImports**: In the example above, the `useContext` import should be cleaned up. When doing `eglot-code-action-organize-imports` it reports `eglot--error: [eglot] No "source.organizeImports" code actions here`. I tried selecting the whole buffer too. the only way I got this working is by placing the cursor on `u` of `useContext` and when doing  `eglot-code-actions` it prompts with `Remove unsused declaration for 'useContext'?`. This works. However ideally I'd do this for all input at once.

Thanks in advance. Again if I missed something obvious then sorry. Really trying eglot for the first time, coming from lsp-mode.

Cheers,
Demis.

     
* Server used:               [typescript-language-server](https://github.com/typescript-language-server/typescript-language-server)
* Emacs version:             29.0.50
* Operating system:          GNU/Guix
* Eglot version:             1.8
* Eglot installation method:       Installed Guix [package](https://guix.gnu.org/packages/emacs-eglot-1.8/), then simply:
[...]

PR #871, opened 2022-03-12
Calculate possible code-actions kinds dynamically

nemethf:
I thought the server must reply with a subset of the client's list of
codeActionKind during the initialization.  But it's clearly not the case.
The typescript-language-server returns these [1]:

  ["source.fixAll.ts" "source.removeUnused.ts" "source.addMissingImports.ts" ...]

The clangd server returns these [2]:

  ["quickfix" "refactor" "info"]

(Additionally, in the current code there is a mismatch between what
Eglot initially sent and the completing read argument of
eglot-code-actions.  "source" and "refactor" is missing from
`eglot-code-actions'. Was this intentional?)

Now, Eglot plays safe, and offers the union of the two lists as
possible completions in `eglot-code-actions'.

[1]: https://github.com/joaotavora/eglot/issues/847#issuecomment-1059760171
[2]: https://github.com/joaotavora/eglot/issues/860

### DIFF
--- a/eglot.el
+++ b/eglot.el
@@ -644,7 +644,7 @@ treated as in `eglot-dbind'."
 
 (cl-defgeneric eglot-client-capabilities (server)
   "What the EGLOT LSP client supports for SERVER."
-  (:method (_s)
+  (:method (server)
            (list
             :workspace (list
                         :applyEdit t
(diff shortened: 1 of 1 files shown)


lljr:
@nemethf I didn't want to create a seperate issue for this to avoid adding noise. But is there a way to Eglot to support https://code.visualstudio.com/docs/languages/jsconfig#_location-of-jsconfigjson . Put simply, to support `jsconfig.json` in multiple locations.

minikN:
Any news on this?

PR #885, opened 2022-03-14
Change dart language server to analysis_server provided with the Dart SDK.

jgoamakf:
The language server currently set for dart is dart_language_server, but it is now deprecated and suggests to use language server included in Dart SDK (https://github.com/natebosch/dart_lsp).

According to the new language server page(https://github.com/dart-lang/sdk/tree/master/pkg/analysis_server/tool/lsp_spec), it accepts two command line options: --client-id and --client-version.

But I don't set these options in this pull request. Because it works and I'm not sure what should be set in this case.


nemethf:
Commit 000598a fixed this.  Thanks.

PR #887, opened 2022-03-15
Add explicit support for latex-mode

manuel-uberti:
I use AUCTeX. Without this change, I get this when doing <kbd>C-x C-f /tmp/test.tex RET</kbd> and <kbd>M-x eglot</kbd>:
```
[eglot] Sorry, couldn't guess for `latex-mode'!
Enter program to execute (or <host>:<port>):
```
With this change, instead, Eglot interacts with `digestif` correctly and connects as expected.


nemethf:
```
 emacs -Q -l ~/.emacs.d/elpa/auctex-13.1.0/auctex-autoloads.el --eval  "(describe-function 'latex-mode)"
```
then in the *Help* buffer:
```
(latex-mode)

  Parent mode: ‘tex-mode’.
```
And Eglot [checks parent modes](https://github.com/joaotavora/eglot/blob/9389d2e4a0976068f129c9b4d20253bc0c562199/eglot.el#L849), so I think this PR is not necessary and something's probably not right in your config.

manuel-uberti:
Thank for the hint, you're right. I removed this line from my configuration:
```emacs-lisp
(push '("\\.tex\\'" . TeX-latex-mode) auto-mode-alist)
```
And everything works as expected. Indeed, this PR is not needed at all.

PR #899, opened 2022-03-27
Implement on-type-formatting support

nemethf:
* eglot.el (eglot--on-type-formatting-chars): New local var, set
and unset ...
(eglot--managed-mode): ... here based on server's
documentOnTypeFormatting capability, ...
(eglot--post-self-insert-hook): and used here to call eglot-format
for on-type-formatting characters.
(eglot-format): Add new optional argument character to request
:textDocument/onTypeFormatting.

* eglot-tests.el (eglot--simulate-key-event): New helper defun.
(rust-on-type-formatting): New test.

### DIFF
--- a/eglot-tests.el
+++ b/eglot-tests.el
@@ -262,6 +262,13 @@ Pass TIMEOUT to `eglot--with-timeout'."
          (eglot-connect-timeout timeout))
     (apply #'eglot--connect (eglot--guess-contact))))
 
+(defun eglot--simulate-key-event (char)
+  "Like (execute-kbd-macro (vector char)), but with `call-interactively'."
+  ;; Also, this is a bit similar to what electric-tests.el does.
+  (setq last-input-event char)
+  (setq last-command-event char)
+  (call-interactively (key-binding (vector char))))
+
 
 ;;; Unit tests
 
@@ -669,6 +676,26 @@ pyls prefers autopep over yafp, despite its README stating the contrary."
       (eglot-format-buffer)
       (should
        (string= (buffer-string) "def a():\n    pass\n\n\ndef b():\n    pass\n")))))
+
+(ert-deftest rust-on-type-formatting ()
+  "Test textDocument/onTypeFormatting agains rust-analyzer."
+  (skip-unless (executable-find "rust-analyzer"))
+  (skip-unless (executable-find "cargo"))
+  (eglot--with-fixture
+      '(("on-type-formatting-project" .
+         (("main.rs" .
+           "fn main() -> () {\n  foo\n    .bar()\n  "))))
+    (with-current-buffer
+        (eglot--find-file-noselect "on-type-formatting-project/main.rs")
+      (let ((eglot-server-programs '((rust-mode . ("rust-analyzer")))))
+        (should (zerop (shell-command "cargo init")))
+        (eglot--sniffing (:server-notifications s-notifs)
+          (should (eglot--tests-connect))
+          (eglot--wait-for (s-notifs 10) (&key method &allow-other-keys)
+             (string= method "textDocument/publishDiagnostics")))
+        (goto-char (point-max))
+        (eglot--simulate-key-event ?.)
+        (should (looking-back "^    \\."))))))
 
 (ert-deftest javascript-basic ()
   "Test basic autocompletion in a JavaScript LSP."
--- a/eglot.el
+++ b/eglot.el
@@ -1974,18 +1974,16 @@ THINGS are either registrations or unregisterations (sic)."
   "If non-nil, value of the last inserted character in buffer.")
 
 (defun eglot--post-self-insert-hook ()
-  "Set `eglot--last-inserted-char', call on-type-formatting if necessary."
+  "Set `eglot--last-inserted-char', maybe call on-type-formatting."
   (setq eglot--last-inserted-char last-input-event)
-  (when (or (eq last-input-event
-                (elt (eglot--server-capable
-                      :documentOnTypeFormattingProvider
-                      :firstTriggerCharacter)
-                     0))
-            (seq-find (lambda (elt) (eq last-input-event (elt elt 0)))
-                      (eglot--server-capable
-                         :documentOnTypeFormattingProvider
-                         :moreTriggerCharacter)))
-    (eglot-format (point) nil (string last-input-event))))
+  (let ((ot-provider (eglot--server-capable :documentOnTypeFormattingProvider)))
+    (when (and ot-provider
+               (or (eq last-input-event
+                       (elt (plist-get ot-provider :firstTriggerCharacter) 0))
+                   (cl-find last-input-event
+                            (plist-get ot-provider :moreTriggerCharacter)
+                            :key #'seq-first)))
+      (eglot-format (point) nil last-input-event))))
 
 (defun eglot--pre-command-hook ()
   "Reset `eglot--last-inserted-char'."
@@ -2373,9 +2371,8 @@ If either BEG or END is nil, format entire buffer.
 Interactively, format active region, or entire buffer if region
 is not active.
 
-If ON-TYPE-FORMAT is non-nil, request on-type-formatting from the
-server.  The argument should be a one-character-long string that
-has just been inserted at BEG."
+If non-nil, ON-TYPE-FORMAT is a character just inserted at BEG
+for which LSP on-type-formatting should be requested."
   (interactive (and (region-active-p) (list (region-beginning) (region-end))))
   (pcase-let ((`(,method ,cap ,args)
                (cond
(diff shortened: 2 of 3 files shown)


eglot.el
```diff
@@ -1974,7 +1989,9 @@ THINGS are either registrations or unregisterations (sic)."
 
 (defun eglot--post-self-insert-hook ()
   "Set `eglot--last-inserted-char'."
```
joaotavora, line 1991: This docstring needs an overhaul, it should become something more generic like `Do Eglot things after a character is inserted.`, maybe with better imagination.

eglot.el
```diff
@@ -2356,14 +2373,22 @@ Try to visit the target file for a richer summary line."
...
+If CHARACTER is non-nil, assume it is an on-type-formatting,
+i.e., the user wants to insert CHARACTER at BEG."
   (interactive (and (region-active-p) (list (region-beginning) (region-end))))
   (pcase-let ((`(,method ,cap ,args)
                (cond
+                ((and beg character)
+                 `(:textDocument/onTypeFormatting
+                   :documentOnTypeFormattingProvider
+                   ,`(:position ,(eglot--pos-to-lsp-position beg)
+                      :ch ,character)))
```
joaotavora, line 2391: You can put the `(string character)` here, or `(string last-input-event)`

eglot.el
```diff
@@ -2356,14 +2373,22 @@ Try to visit the target file for a richer summary line."
...
 
-(defun eglot-format (&optional beg end)
+(defun eglot-format (&optional beg end character)
   "Format region BEG END.
 If either BEG or END is nil, format entire buffer.
 Interactively, format active region, or entire buffer if region
-is not active."
+is not active.
+
+If CHARACTER is non-nil, assume it is an on-type-formatting,
```
joaotavora, line 2382: Say the type of `CHARACTER`, i.e. that it is a character (or a one-character string).  Perhaps name it `ON-TYPE-FORMAT`.

eglot.el
```diff
@@ -1651,7 +1665,8 @@ Use `eglot-managed-p' to determine if current buffer is managed.")
               (delq (current-buffer) (eglot--managed-buffers server)))
         (when (and eglot-autoshutdown
                    (null (eglot--managed-buffers server)))
-          (eglot-shutdown server))))))
+          (eglot-shutdown server))))
+    (setq eglot--on-type-formatting-chars nil)))
```
joaotavora, line 1669: If this were a server slot you wouldn't need to reset it.

eglot.el
```diff
@@ -1623,7 +1626,18 @@ Use `eglot-managed-p' to determine if current buffer is managed.")
                     #'eglot-imenu))
     (unless (eglot--stay-out-of-p 'flymake) (flymake-mode 1))
     (unless (eglot--stay-out-of-p 'eldoc) (eldoc-mode 1))
-    (cl-pushnew (current-buffer) (eglot--managed-buffers (eglot-current-server))))
+    (cl-pushnew (current-buffer) (eglot--managed-buffers (eglot-current-server)))
+    (setq eglot--on-type-formatting-chars
```
joaotavora, line 1630: I think this should be saved in a server slot somewhere inside the `:success-fn` callback of `eglot-connect`.

eglot.el
```diff
@@ -1623,7 +1626,18 @@ Use `eglot-managed-p' to determine if current buffer is managed.")
...
+          (cl-loop
+           for char in (append
```
joaotavora, line 1632: I think you just just use `cons` here.  And maybe avoid the loop completely and just store the list of 1-char strings in the slot.  You can convert back to char when doing the test, I don't think it's expensive.

eglot.el
```diff
@@ -1974,7 +1989,9 @@ THINGS are either registrations or unregisterations (sic)."
 
 (defun eglot--post-self-insert-hook ()
   "Set `eglot--last-inserted-char'."
-  (setq eglot--last-inserted-char last-input-event))
+  (setq eglot--last-inserted-char last-input-event)
+  (when (memq last-input-event eglot--on-type-formatting-chars)
```
joaotavora, line 1993: I would use `cl-find` with a `:key` that extracts the char.  Putting more of my ideas together, this would be
```lisp
(when-let* ((server (eglot-current-server))
            (char (member (string last-input-event)
                          (eglot--on-type-formatting server))))
  (eglot-format beg nil char))
```

eglot-tests.el
```diff
@@ -670,6 +677,29 @@ pyls prefers autopep over yafp, despite its README stating the contrary."
...
+           "fn main() -> () {\n  foo\n    .bar()\n  "))))
+    (with-current-buffer
+        (eglot--find-file-noselect "on-type-formatting-project/main.rs")
+      (let ((eglot-server-programs '((rust-mode . ("rust-analyzer")))))
+        (should (zerop (shell-command "cargo init")))
+        (eglot--sniffing (:server-notifications s-notifs)
+          (should (eglot--tests-connect))
+          ;; Idle until the server finishes its housekeeping by
+          ;; waiting for publishDiagnostics.  This notification
+          ;; arrives quite late.  experimental/serverStatus arrives
```
joaotavora, line 695: Then maybe wait for `experimental/serverStatus`?  Is it too unstable, you think?

eglot-tests.el
```diff
@@ -262,6 +262,13 @@ Pass TIMEOUT to `eglot--with-timeout'."
          (eglot-connect-timeout timeout))
     (apply #'eglot--connect (eglot--guess-contact))))
 
+(defun eglot--simulate-key-event (char)
```
joaotavora, line 265: There is `ert-simulate-keys` in `ert-x.el`.  A hack it itself seems less hacky than this (and not our responsibility ;-) )  Doesn't it work?  If it doesn't then fine to add this substitute.

eglot.el
```diff
@@ -1548,6 +1548,9 @@ and just return it.  PROMPT shouldn't end with a question mark."
 (defvar-local eglot--saved-bindings nil
   "Bindings saved by `eglot--setq-saving'.")
 
+(defvar-local eglot--on-type-formatting-chars nil
```
joaotavora, line 1551: Make this a server slot. If nothing else, it'll simplify things when we support multiple servers.  _if_ we ever support that :-)

joaotavora (commented):
Thanks very much! Looks good, but please consider my comments and changes.

joaotavora:
Actually, I see that even the new server slot isn't needed.  This is already in the server as "capabilities".  So just grab it from there.  I don't think processing required is extremely more onerous.  At least not when compared to the performance hit we're already incurring in when calling `jsonrpc-request` from the `post-self-insert-hook`.

So this should simplify the code a lot.

eglot-tests.el
nemethf: `ert-simulate-keys` is not in Emacs 27.1, but more importantly it does something else: it provides user input for a function if it queries the user input (eg. read-from-minibuffer) during its execution.  The goal of `eglot--simulate-key-event` is initiate a command-loop for a given keypress.  I think the whole thing can be replaced with `(setq last-input-event ?.) (self-insert-command 1 ?.)` as that also calls post-self-insert-hooks.  But then some part of the implementation becomes part of the test.  An earlier iteration of this PR redefined keybindings for the shake of execution speed.  eglot--simulate-key-event worked with that implementation as well.

Shall I (a) change the test to use `(setq last-input-event ?.) (self-insert-command 1 ?.)`, or (b) extend the documentation of  `eglot--simulate-key-event`?

eglot-tests.el
nemethf: It is stable, but the server only sends it if the client requires it.  I implemented this [here](https://github.com/nemethf/eglot-x/blob/master/eglot-x.el#L189-L193), but I though that's too intrusive for eglot-test.el.  Shall I copy that here?

nemethf:
> Actually, I see that even the new server slot isn't needed. This is already in the server as "capabilities". So just grab it from there. I don't think processing required is extremely more onerous. At least not when compared to the performance hit we're already incurring in when calling `jsonrpc-request` from the `post-self-insert-hook`.
> 
> So this should simplify the code a lot.

The goal of precomputing eglot--on-type-formatting-chars is make post-self-insert-hook run fast for chars that are not trigger chars for on-type-formatting.  Shall I (a) add this to the documentation of eglot--on-type-formatting-chars, or (b) simplify the code and calculate the list every time the user presses a key?

I'll address the rest of your comments once I see clear with these two questions.

Thanks for the review.

joaotavora:
Please use (b) and use cons to make list construction easier.

João

eglot.el
```diff
@@ -1973,8 +1973,18 @@ THINGS are either registrations or unregisterations (sic)."
   "If non-nil, value of the last inserted character in buffer.")
 
 (defun eglot--post-self-insert-hook ()
-  "Set `eglot--last-inserted-char'."
-  (setq eglot--last-inserted-char last-input-event))
+  "Set `eglot--last-inserted-char', call on-type-formatting if necessary."
+  (setq eglot--last-inserted-char last-input-event)
+  (when (or (eq last-input-event
+                (aref (eglot--server-capable
```
joaotavora, line 1979: won't this `aref` :boom: if the server is not capable of this?

nemethf:
I've force-pushed a new version addressing your concerns including the aref :boom:.

It turned out the experimental/serverStatus is just as slow a waiting for the first diagnostic notification.

joaotavora:
I took the liberty of making some minor tweaks and pushed this.  Let's keep eyes peeled for any problems (performance, :boom:, etc), though I think this is pretty sane.

nemethf:
Thanks!

PR #1033, opened 2022-09-10
Allow eglot-workspace-configuration to be a plist

astoff:
I would also suggest changing the documentation so it refers only to plists, or only to alists. The current formulation is pretty confusing, since it mixes both things. Also, I guess we shouldn't make reference to alists with keyword or string keys, which are unusual and not accepted by `json-serialize` respectively.

So, in https://github.com/joaotavora/eglot#simple-eglot-workspace-configuration, either

```
((python-mode
  . ((eglot-workspace-configuration
      . ((pyls . (plugins . (jedi_completion . (include_params . t)))))))))
```
or

```
((python-mode
  . ((eglot-workspace-configuration
      . (:pyls (:plugins (:jedi_completion (:include_params t))))))))
```

### DIFF
--- a/eglot.el
+++ b/eglot.el
@@ -2223,12 +2223,15 @@ When called interactively, use the currently active server"
    server :workspace/didChangeConfiguration
    (list
     :settings
-    (or (cl-loop for (section . v) in (eglot--workspace-configuration server)
-                 collect (if (keywordp section)
-                             section
-                           (intern (format ":%s" section)))
-                 collect v)
-        eglot--{}))))
+    (let ((config (eglot--workspace-configuration server)))
+      (or (and (consp (car config))
+               (cl-loop for (section . v) in config
+                        collect (if (keywordp section)
+                                    section
+                                  (intern (format ":%s" section)))
+                        collect v))
+          config
+          eglot--{})))))
 
 (cl-defmethod eglot-handle-request
   (server (_method (eql workspace/configuration)) &key items)
@@ -2245,14 +2248,19 @@ When called interactively, use the currently active server"
                          (project-root (eglot--project server)))))
                 (setq-local major-mode (eglot--major-mode server))
                 (hack-dir-local-variables-non-file-buffer)
-                (alist-get section (eglot--workspace-configuration server)
-                           nil nil
-                           (lambda (wsection section)
-                             (string=
-                              (if (keywordp wsection)
-                                  (substring (symbol-name wsection) 1)
-                                wsection)
-                              section))))))
+                (let ((config (eglot--workspace-configuration server)))
+                  (alist-get section
+                             (if (symbolp (car config))
+                                 (cl-loop for (a b) on config by #'cddr
+                                          collect (list a b))
+                               config)
+                             nil nil
+                             (lambda (wsection section)
+                               (string=
+                                (if (keywordp wsection)
+                                    (substring (symbol-name wsection) 1)
+                                  wsection)
+                                section)))))))
           items)))
 
 (defun eglot--signal-textDocument/didChange ()


fbergroth:
I was recently reminded of a closed PR of mine, that does something similar; #790.
It dumps the config recursively and tries to fix inconsistencies between `json-encode` and `json-serialize`.
Since you also looked into this and I vaguely remember the details, do you think it covers this, and should be merged?

joaotavora:
> I was recently reminded of a closed PR of mine, that does something similar; https://github.com/joaotavora/eglot/pull/790.

Yes, please merge the two PRs, explain clearly:

* what problem this solves
* if there is any backward incompatibility
* what the changes to the documentation should be

astoff:
To be honest, I'm not a fan of attempting to fix inconsistencies of the two encoders, which AFAICT are pretty mild. Specifically, the C encoder gives an error on alists and plists where the keys are strings, and it doesn't remove the leading colon if you use a keyword as a key in an alist.

I think these problems are best solved by by restricting to the usual constructs in the examples and docstrings: alists with plain symbol keys and plists with keyword keys.

If I were to attempt something smart here, I would rather allow string keys for a "flattened out" representation of deeply nested structures. So, for instance, you could write
```
((python-mode
  . ((eglot-workspace-configuration
      "pyls.plugins.jedi_completion.include_params" t
      "pyls.plugins.something_else" [1 2 3]
```
and this would be equivalent to saying
```
((python-mode
  . ((eglot-workspace-configuration
      . ((:pyls . (:plugins (:jedi_completion (:include_params t))))))))
      . ((:pyls (:plugins (:jedi_completion (:include_params . t)) :something_else [1 2 3]))))))
```

My tiny merge request here is compatible with this future possiblity. How about #790?

joaotavora:
Can you explain why these incompatibilities cannot or should not be addressed in jsonrpc.el?

joaotavora:
I still struggle to understand what the problem is.

 Can you state again, for my benefit, exactly what the user may type today that leads to different behaviour according to the user's (unknown and underlying) use of json.el or json.c? 

Thereafter state which of these two behaviours you think is the correct one.

astoff:
> Can you state again, for my benefit, exactly what the user may type today that leads to different behaviour according to the user's (unknown and underlying) use of json.el or json.c?

Okay, but this is more related to #790, so I've replied there.

astoff:
The purpose of this PR is pretty restricted: to allow both of the following forms:

```
((python-mode
  . ((eglot-workspace-configuration
      . ((pyls . (plugins . (jedi_completion . (include_params . t)))))))))
```
or
```
((python-mode
  . ((eglot-workspace-configuration
      . (:pyls (:plugins (:jedi_completion (:include_params t))))))))
```

Currently only the first one is accepted.

joaotavora:
ok thanks for the clarification.

astoff:
If you like the suggestion of https://github.com/joaotavora/eglot/pull/1033#issuecomment-1243304059, it would be better to implement it directly instead of merging this.

fbergroth:
It seems this assumes configuration to be an alist? https://github.com/joaotavora/eglot/blob/06e6dd6693123a737fb883dcf277432d63d6518e/eglot.el#L2248

joaotavora:
Yes, whatever the code does now is behaviour to be kept. We can of course
enhance with backward-compatible things.

astoff:
Okay, I missed the case of the `workspace.configuration` server -> client request. So this change, as is right now, would break this functionality.  With that said:

1. The server submits a request to the client? That's madness. 
2. The "specification" of the `workspace.configuration` method (https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_configuration) is totally vague. I guess the expected behavior here is "whatever VS Code does".
3. But if I'm guessing correctly, the `section` argument can be a nested thingy like `"pyls.plugins.jedi"`, and if that's correct then the current implementation is not  conformant.


In my opinion Eglot should not support this method. The `initializationOptions` and the `workspace.didChangeConfiguration` method should suffice to configure the server. WDYT?

joaotavora:
I'm not opposed to this PR and I think it'd be nice to allow `eglot-workspace-configuration` to be simply a plist, even though I think that the latest doc update makes things simple.

But it's missing this bit as pointed out earlier (https://github.com/joaotavora/eglot/pull/1033#issuecomment-1249623092).  

Also I'd prefer if the change is small.  If it becomes too big/complex, I don't think it's worth it.

astoff:
> But it's missing this bit as pointed out earlier ([#1033 (comment)](https://github.com/joaotavora/eglot/pull/1033#issuecomment-1249623092)).

That comment points to a separate issue. Did you look at the [LSP spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_configuration), where it gives `"cpp.formatterOptions"` as a possible value of `"section"` in the `workspace/configuration` request? Do you agree this is already not handled correctly by Eglot?

There is an interdependency between that issue and this PR. From my perspective it would be better to fix the issue _after_ merging this PR, since only then it will be clear what are all the forms that `eglot-workspace-configuration` can have.

joaotavora:
> Do you agree this is already not handled correctly by Eglot?

I don't know.  I just know that if I merge your pr and tell people they can use plists, that piece of code that was added in 2019 to fix, presumably, some real problem, will cease to work.

> There is an interdependency between that issue and this PR. From my perspective it would be better to fix the issue after merging this PR, since only then it will be clear what are all the forms that eglot-workspace-configuration can have.

I'm sorry, I don't view things that way. Allowing `e-w-configuration` to be a plist is a nice-to-have, not a bugfix.  So I don't want to break things over it.

astoff:
Well, I've pushed a change to the `workspace/configuration` handler, but I can't test it because I have never seen this request happen. So I suspect this pr will not advance, but we'll see.

eglot.el
```diff
@@ -2223,12 +2223,14 @@ When called interactively, use the currently active server"
...
    (list
     :settings
-    (or (cl-loop for (section . v) in (eglot--workspace-configuration server)
-                 collect (if (keywordp section)
-                             section
-                           (intern (format ":%s" section)))
-                 collect v)
-        eglot--{}))))
+    (let ((config (eglot--workspace-configuration server)))
+      (if (consp (car-safe config))
```
joaotavora, line 2227: Just `car` is fine.  `car-safe` is for non-lists (vectors, symbols, etc), which config never is here.

eglot.el
```diff
@@ -2245,14 +2247,21 @@ When called interactively, use the currently active server"
...
                 (hack-dir-local-variables-non-file-buffer)
-                (alist-get section (eglot--workspace-configuration server)
-                           nil nil
-                           (lambda (wsection section)
-                             (string=
-                              (if (keywordp wsection)
-                                  (substring (symbol-name wsection) 1)
-                                wsection)
-                              section))))))
+                (let ((conf (eglot--workspace-configuration server)))
```
joaotavora, line 2250: Wonder if we can use `plist-get` for the plist case as the reciprocal of `alist-get` for the alist case and simply reuse the predicate.   

Could be less complex than converting the plist to alist, though that is not a bad idea either.

 BTW Converting a plist to an alist is most idiomatically done like this in Common Lisp's loop:

```
(cl-loop for (a b) on 'plist by #'cddr collect (list a b))
```

eglot.el
joaotavora: I think this whole block is neater like:
```
(or (and (consp (car config))
              alist-stuff
    config
    eglot--{})
```

joaotavora (changes requested):
Of course, we also need the doc changes that suggest plists only in README.md.  And ideally tests.

eglot.el
astoff: Okay, I admit I was getting lazy there. Since you seem convinced that this can be merged in some form, I'll improve it using one of your suggestions.

astoff: Sadly, the calling conventions are reversed:
```
(plist-get '(0 a 1 b 2 c) 1 '<) ==> c
(alist-get 1 '((0 . a) (1 . b) (2 . c)) nil nil '<) ==> a
```
so the easiest way around it is to convert the plist to an alist.

eglot.el
astoff: Okay, done.

eglot.el
joaotavora: Yes, that's fine.  Or even simpler, always use a plist.  I.e. make use of the existing `eglot--workspace-configuration-plist`.  Then change that function to be a noop if its argument is already a plist.

astoff: > eglot--workspace-configuration-plist

You know, instead of all these ad-hoc conversions I think we would be better off with 1 helper function that looks up a key in a mapping and doesn't care whether the key is a string, keyword or normal symbol and whether the mapping is an alist or plist.

Also in the docs the configurations shouldn't use keywords. Alists and plists of normal symbols work just fine in that context.

joaotavora: Sorry, no.  I don't want to overcomplicate this. I just want to allow `eglot-workspace-configuration` to be a plist, which is the title of this PR.  As far as I can tell, this diff is all there is to it:
```diff
diff --git a/eglot.el b/eglot.el
index ff94d5ca5f..459a619311 100644
--- a/eglot.el
+++ b/eglot.el
@@ -2232,11 +2232,14 @@ format described above.")
 
 (defun eglot--workspace-configuration-plist (server)
   "Returns `eglot-workspace-configuraiton' suitable serialization."
-  (or (cl-loop for (section . v) in (eglot--workspace-configuration server)
-               collect (if (keywordp section) section
-                         (intern (format ":%s" section)))
-               collect v)
-      eglot--{}))
+  (let ((val (eglot--workspace-configuration server)))
+    (or (and (consp (car val))
+             (cl-loop for (section . v) in val
+                      collect (if (keywordp section) section
+                                (intern (format ":%s" section)))
+                      collect v))
+        val
+        eglot--{})))
 
 (defun eglot-signal-didChangeConfiguration (server)
   "Send a `:workspace/didChangeConfiguration' signal to SERVER.
@@ -2263,9 +2266,8 @@ When called interactively, use the currently active server"
                          (project-root (eglot--project server)))))
                 (setq-local major-mode (eglot--major-mode server))
                 (hack-dir-local-variables-non-file-buffer)
-                (alist-get section (eglot--workspace-configuration server)
-                           nil nil
-                           (lambda (wsection section)
+                (plist-get (eglot--workspace-configuration-plist server) section
+                           (lambda (section wsection)
                              (string=
                               (if (keywordp wsection)
                                   (substring (symbol-name wsection) 1)
```

astoff:
Fine, so we can close this ticket. But I doubt that https://github.com/joaotavora/eglot/commit/e4b3a5315803756f56292a04fb0b2065d6fc3e03 works. You defined a function `eglot--workspace-configuration-plist` that returns a plist _or a hash table_, and then you call `plist-get` on its output.

I also think you would save work if you had a function that normalizes the config to have non-keyword keys, so you can call `plist-get` or `alist-get` with a simple `string=` as test function. But that's just a _palpite_ (hard to translate precisely, that word).

joaotavora:
>You defined a function eglot--workspace-configuration-plist that returns a plist or a hash table, and then you call plist-get on its output.

Hmm good point. I missed that, I guess I can just make it return nil

> But that's just a palpite (hard to translate precisely, that word).

That would be "hunch" I think.  But palpite if much better of course.

PR #1055, opened 2022-09-21
eglot-server-programs: Add support for nil

leungbk:
* eglot.el (eglot-server-programs): Add nil

* README.md: Mention nil

* NEWS.md: Mention nil

This is more featureful and more actively maintained than `rnix-lsp`,
so I have placed it first in `eglot-alternatives`.


bhankas:
Another `nil` user here, can confirm it is much more responsive and almost useful compared to `rnix-lsp` which I never quite got to being useful in any nix buffer ever.

leungbk:

PR #1082, opened 2022-10-16
Add `ltex-ls` to supported servers

abougouffa:
First, thank you so much for this beautiful package, I moved to Eglot some time ago, and I'm really surprised by its power yet simplicity, congratulations on that!

This PR adds [ltex-ls](https://github.com/valentjn/ltex-ls) support, which is is a server that provides LanguageTool integration via LSP, which gives advanced natural languages checking (grammar and so on) for text documents. 

This PR:

- Add `ltex-ls` for relevant modes, as a second choice for existing ones (Latex related and Markdown).
- Mention it in the README, in the relevant modes.

LTeX supports Org mode, LaTeX, reStructuredText (... and others), and so, do not give much false positives as it understands the markup language's structure.

I'm using it regularly, and it works fine with Eglot, it can be customized using the `eglot-workspace-configuration`, for example, by adding this in the `dir-locals.el` file:

```elisp
((nil (eglot-workspace-configuration
       . ((ltex . ((language . "fr")
                   (disabledRules . ((fr . ["FRENCH_WHITESPACE"])))
                   (additionalRules . ((languageModel . "/usr/share/ngrams/")))))))))
```

More information about the server options and configuration options can be found at https://valentjn.github.io/ltex

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,20 +60,22 @@ find-library` can help you tell if that happened.
 * Javascript's [TS & JS Language Server][typescript-language-server]
 * Kotlin's [kotlin-language-server][kotlin-language-server]
 * Lua's [lua-lsp][lua-lsp]
-* Markdown's [marksman][marksman]
+* Markdown's [marksman][marksman] or [ltex-ls][ltex-ls]
 * Mint's [mint-ls][mint-ls]
 * Nix's [rnix-lsp][rnix-lsp]
 * Ocaml's [ocaml-lsp][ocaml-lsp]
+* Org's [ltex-ls][ltex-ls]
 * Perl's [Perl::LanguageServer][perl-language-server]
 * PHP's [php-language-server][php-language-server]
 * PureScript's [purescript-language-server][purescript-language-server]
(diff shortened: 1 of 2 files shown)


abougouffa:
Yep, they do different things, is there a way to tell Eglot to use two servers at the same time?

doolio:
> In the multi-server situation expect I expect some server to remain the "primary server", which will be where requests like textDocument/rename end up.

This is seemingly possible with the [`pyright` ](https://github.com/microsoft/pyright/blob/main/docs/settings.md) server which has the following option allowing it to be used with other servers. I've not tried this yet though.

> **pyright.disableLanguageServices** [boolean]: Disables all language services except for “hover”. This includes type completion, signature completion, find definition, find references, and find symbols in file. This option is useful if you want to use pyright only as a type checker but want to run another Python language server for language service features.

abougouffa:
Thanks for the clarification @joaotavora , 
In this case, I think the commit is OK, for LaTeX and Markdown, the `ltex-ls` is provided as an alternative, as it doesn't provide checking for the markup language but for the natural language (grammar checking). I think it can be provided as an alternative, in all cases, people will likely enable it on demand while checking a document for mistakes, as `ltex-ls` can be too heavy to run all the time.

PR #1346, opened 2023-12-31
Add nushell language server

nibon7:
nvim-lspconfig and helix have LSP support for [Nushell](https://github.com/nushell/nushell), it would be nice if eglot could support it natively. 

https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/server_configurations/nushell.lua
https://github.com/helix-editor/helix/blob/85fce2f5b6c9f35ab9d3361f3933288a28db83d4/languages.toml#L51

lsp


nibon7:
Merged into emacs-29.

https://git.savannah.gnu.org/cgit/emacs.git/commit/?h=emacs-29&id=a3987127618b9fe49b88807f0268ec9abcc7396f

PR #1370, opened 2024-03-07
Avoid message-spamming when jit-lock-context-unfontify-pos nil

unhammer:
In js-mode, I get an error with the message

      (jit-lock--run-functions 3698 3699) [UNPR HOST api.js]: Wrong type argument: number-or-marker-p, nil

for almost every edit. The nil-check in this patch fixes it. The variable is by default nil and docs say it can be nil if contextual fontification is disabled.


joaotavora:
A similar fix has been upstream into Emacs master.

PR #1371, opened 2024-03-12
remove rnix-lsp, link nil and nixd

adamcstephens:
rnix-lsp is an archived project as of January. It can no longer be securely installed due to dependence on nix 2.15, which has been marked insecure. Finally, I've merged the removal of rnix-lsp from nixpkgs. NixOS/nixpkgs#294488


adamcstephens:
Ok, well consider it a suggestion then. I won't be contributing this to emacs core.

PR #1421, opened 2024-07-03
support nix-ts-mode

acid-bong:


acid-bong:
Whoops, didn't read the readme section

PR #1482, opened 2025-01-27
Add camel-cased "OmniSharp" as C# LSP server

DamienCassou:
Some distributions provide a "OmniSharp" binary and no "omnisharp". This is for example the case for Nixpkgs:
https://github.com/NixOS/nixpkgs/blob/9f023565a4ecb13a6863f72c05e598c801300b9e/pkgs/by-name/om/omnisharp-roslyn/package.nix#L121

### DIFF
--- a/eglot.el
+++ b/eglot.el
@@ -325,6 +325,7 @@ automatically)."
     ((csharp-mode csharp-ts-mode)
      . ,(eglot-alternatives
          '(("omnisharp" "-lsp")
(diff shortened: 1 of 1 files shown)


joaotavora:
Thanks, but this is not the place to submit patches.  This repo is not the upstream anymore.  Send them to bug-gnu-emacs@gnu.org.  Suggest replacing the `omnisharp` entry with `OmniSharp` (though I really don't care that much).

PR #1524, opened 2025-06-06
readme: add ghdl-language-server for vhdl.

csantosb:
Ghdl is a simulator, synthesis and lsp for vhdl.

It is compatible with eglot.


joaotavora:
Read readme

joaotavora:
This is not the upstream anymore. Submit patch to Emacs

PR #1553, opened 2025-10-31
eglot.el (eglot-rename): Pre-fill prompt with current symbol

zarkone:
Fix an inconvenience I often stumble upon 😊


joaotavora:
Ok, ok it's a decent idea, but submit this as a patch to the upstream. See the README

Issue #661, opened 2021-04-05
Find definition won't work for external dependencies in Clojure


Andre0991:
Hi. I'm experimenting with using `eglot` with Clojure.

I've had no issues so far except that `xref-find-definitions` won't work for external dependencies.
When I use it on some symbol that refers to an external lib, Emacs asks me if I want to create the directory:

```
Directory `/Users/andreperictavares/dev/peric/clj-new/src/clj_new/file:/Users/andreperictavares/.m2/repository/org/clojure/clojure/1.10.1/clojure-1.10.1.jar!/clojure/java/' does not exist! Create it? (y or n) y
```

The expected behaviour would be to open a buffer with the contents of the file in the jar (and manage it in the same project? Not sure).

Here's the an eglot log sample for this issue:

```
[client-request] (id:7) Sun Apr  4 21:44:24 2021:
(:jsonrpc "2.0" :id 7 :method "textDocument/definition" :params
          (:textDocument
           (:uri "file:///Users/andreperictavares/dev/nu/luna/src/luna/logic/error.clj")
           :position
           (:line 4 :character 23)))
[server-reply] (id:7) Sun Apr  4 21:44:25 2021:
(:jsonrpc "2.0" :id 7 :result
          (:uri "jar:file:////Users/andreperictavares/.m2/repository/prismatic/schema/1.1.9/schema-1.1.9.jar!/schema/core.clj" :range
                (:start
                 (:line 0 :character 4)
                 :end
                 (:line 0 :character 15))))
```

## Steps to reproduce
0. [Download](https://clojure.org/guides/getting_started#_clojure_installer_and_cli_tools) the `clojure` binary

1. Download the native clojure-lsp binary in https://github.com/clojure-lsp/clojure-lsp/releases

2. Configure `eglot` to use it
```
(add-to-list 'eglot-server-programs '(clojure-mode . ("clojure-lsp")))
```

3. Clone `clj-new` and install its dependencies.
```
git clone https://github.com/seancorfield/clj-new.git
clj
```

4. Open the project in `Emacs`, go to `clj-new/src/clj/generate/def.clj` and use `xref-find-definitions` on some external lib like `clojure.java.io` ([line 5](https://github.com/seancorfield/clj-new/blob/a14ba8256d9a0e1d659691be211007ea6eb102b2/src/clj/generate/def.clj#L5) in this file).

[...]

PR #1583, opened 2026-03-26
Don't mangle non-file URIs with expand-file-name

BrianHotopp:
## Summary

When `eglot-uri-to-path` returns a non-file URI unchanged (e.g. `jar:file:///path.jar!/Foo.scala` from Metals, clojure-lsp, jdtls, etc.), three call sites pass it to `expand-file-name`, which treats the URI as a relative path and prepends `default-directory`:

```
jar:file:///home/user/.m2/foo.jar!/com/Bar.scala
→ /project/dir/jar:file:/home/user/.m2/foo.jar!/com/Bar.scala
```

This is inconsistent with the approach established in bug#58790: eglot passes non-file URIs through for `file-name-handler-alist` handlers (e.g. [jarchive](https://elpa.gnu.org/packages/jarchive.html)) to deal with. The bug is that eglot's own code mangles the URIs before they ever reach any handler.

### Fix

Add `eglot--uri-p` to detect non-file URIs (by checking if `url-type` returns a scheme longer than 1 character, to exclude Windows drive letters), and guard three `expand-file-name` call sites:

- `eglot--xref-make-match` — hash store into `eglot--servers-by-xrefed-file`
- `eglot-current-server` — hash lookup from `eglot--servers-by-xrefed-file`
- `eglot--flymake-handle-push` — path for buffer visiting + diagnostics

### Observed symptoms

With Metals (Scala LSP), following `xref-find-definitions` to a symbol in a dependency source jar:

1. The opened buffer gets a mangled `buffer-file-name` with the project dir prepended to the `jar:file:` URI
2. Eglot sends `textDocument/didOpen` with a fabricated `file:///...jar%3Afile%3A/...` URI
3. The language server crashes with `NoSuchFileException` for the bogus path
4. `eglot-extend-to-xref` silently fails because the hash key stored by `eglot--xref-make-match` (mangled) doesn't match the lookup key in `eglot-current-server`

### Related

- #661 — original discussion of jar:file: URI handling
- bug#58790 — the fix that made `eglot-uri-to-path` pass non-file URIs through

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/eglot.el
+++ b/eglot.el
@@ -1239,6 +1239,16 @@ object."
           (concat remote-prefix normalized))
       uri)))
 
+(defsubst eglot--uri-p (path)
+  "Return non-nil if PATH is a URI rather than a file-system path.
+Detects non-file URIs returned unchanged by `eglot-uri-to-path',
+such as \"jar:file:///...\".  These must not be passed to
+`expand-file-name', which would mangle them into bogus local paths."
+  (let ((scheme (url-type (url-generic-parse-url path))))
+    ;; Real URI schemes are always longer than 1 character.
+    ;; A single-character "scheme" is a Windows drive letter (e.g. C:).
+    (and scheme (> (length scheme) 1))))
+
 (cl-defun eglot-path-to-uri (path &key truenamep)
   "Convert PATH, a file name, to LSP URI string and return it.
 TRUENAMEP indicated PATH is already a truename."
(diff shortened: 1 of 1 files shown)


BrianHotopp:
Closing this — the `expand-file-name` calls I patched here are real code smells but don't cause the user-visible crash. The crash happens inside `find-file-noselect` (in xref.el), which is correctly intercepted by a `file-name-handler-alist` entry from the [jarchive](https://elpa.gnu.org/packages/jarchive.html) package. The existing eglot design (pass non-file URIs through for handlers to deal with) is working as intended. Sorry for the noise.

joaotavora:
Yeah, LLMs are a bit too eager sometimes

joaotavora:
But I might need something similar in the future. One user is requesting support for a special "untitled:" URL scheme to create a temporary buffer...

BrianHotopp:
Closing this again on my end — the patched `expand-file-name` calls don't fix the actual user-visible crash (which jarchive's `file-name-handler-alist` entry already handles correctly). If the `untitled:` scheme work ends up needing similar URI-preservation handling, feel free to cherry-pick from here or ping me and I'll rebase. Thanks for taking a look.

